PR #35, opened 2022-01-13
Feedback wanted: API examples for distributed execution

skrawcz:
This is a PR for illustrative purposes only. The point here it so try to settle on an API direction for enabling distributed execution. See last commit.

## Additions

- Three files with examples on how one might run hamilton on Dask.

## Removals

- N/A

## Changes

- N/A

## Testing

1. N/A

## Screenshots
N/A

## Notes
N/A

## Todos

- [ ] get feedback on the contents here.

## Checklist

- [X] PR has an informative and human-readable title
~- [ ] Code can be automatically merged (no conflicts)~
~- [ ] Code follows the standards laid out in the [TODO link to standards]()~
~- [ ] Passes all existing automated tests~
~- [ ] Any _change_ in functionality is tested~
~- [ ] New functions are documented (with a description, list of inputs, and expected output)~
- [X] Placeholder code is flagged / future todos are captured in comments
~- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)~
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist
N/A

### DIFF
--- a/examples/dask/example_one/run_dask_two.py
+++ b/examples/dask/example_one/run_dask_two.py
@@ -1,0 +1,75 @@
+import importlib
+import logging
+import sys
+
+import pandas as pd
+from hamilton import driver, dask_executor, base
+
+from dask import dataframe
+
+logger = logging.getLogger(__name__)
+
+
+if __name__ == '__main__':
+    module_name = 'my_functions'
+
+    from dask.distributed import Client, LocalCluster
+    # Setup a local cluster.
+    # By default this sets up 1 worker per core
+    cluster = LocalCluster()
+    client = Client(cluster)
+    print(client.cluster)
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+    module = importlib.import_module(module_name)
+    initial_columns = {  # load from actuals or wherever -- this is our initial data we use as input.
+        'signups': pd.Series([1, 10, 50, 100, 200, 400]),
+        'spend': pd.Series([10, 10, 20, 40, 40, 50]),
+    }
+
+    """Option 2(a):
+    
+    1. People instantiate a specific Driver.
+    2. The Driver can contain as much or as little logic as it wants, e.g. it can convert input sets into the dask
+    data type.
+    3. The Driver dictates the return type and is fixed. E.g. only returns pandas DF. 
+    """
+    dask_parameters = {
+        'default_npartitions': 2,
+        'convert_to_dask_data_type': True
+    }
+    dr = dask_executor.DaskDriver(
+        initial_columns,
+        module,
+        dask_client=client,
+        dask_parameters=dask_parameters)
+
+    """Option 2(b):
+
+    1. People instantiate a specific Driver.
+    2. The Driver can contain as much or as little logic as it wants, e.g. it can convert input sets into the dask
+    data type.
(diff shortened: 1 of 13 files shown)


skrawcz:
@chmp ^

skrawcz:
(also I'm confused why @chmp's commits are showing up here? They should already be on the distributed branch...)

examples/dask/example_one/run_dask_two.py
```diff
@@ -0,0 +1,75 @@
...
+    print(client.cluster)
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+    module = importlib.import_module(module_name)
+    initial_columns = {  # load from actuals or wherever -- this is our initial data we use as input.
+        'signups': pd.Series([1, 10, 50, 100, 200, 400]),
+        'spend': pd.Series([10, 10, 20, 40, 40, 50]),
+    }
+
+    """Option 2(a):
```
elijahbenizzy, line 30: This is my vote (if I understand everything correctly). 

Let's think about separation of responsibility:

1. FunctionGraph specifies the shape of the DAG
2. Executor executes it
3. Driver makes it easy to use

Thus, if we want to...
- Change the function graph shape we modify the function graph
- Change the way its executed we modify the executor
- Change the API for an end-user we change the Driver
The benefit of this is the driver can abstract any logic and we're not opinionated about it. BUT, we can build higher-level abstractions in there over time. Same with the results_builder -- for now I think we want to just return a `dict[str, Any]`, we can always add a parameter, and I'm not convinced that results building will be unified across different executors/drivers...

Need to mull over results building a little more though.

hamilton/driver.py
```diff
@@ -32,14 +32,18 @@ class Variable:
...
-    def __init__(self, config: Dict[str, Any], *modules: ModuleType):
+    def __init__(self, config: Dict[str, Any], *modules: ModuleType, executor=None):
         """Constructor: creates a DAG given the configuration & modules to crawl.
 
         :param config: This is a dictionary of initial data & configuration.
                        The contents are used to help create the DAG.
         :param modules: Python module objects you want to inspect for Hamilton Functions.
         """
-        self.graph = graph.FunctionGraph(*modules, config=config)
+        if executor is None:
```
elijahbenizzy, line 42: Executor should not be mutable (IMO) so we should be able to pass it in as the default

hamilton/driver.py
```diff
@@ -32,14 +32,18 @@ class Variable:
 class Driver(object):
     """This class orchestrates creating and executing the DAG to create a dataframe."""
 
-    def __init__(self, config: Dict[str, Any], *modules: ModuleType):
+    def __init__(self, config: Dict[str, Any], *modules: ModuleType, executor=None):
```
elijahbenizzy, line 35: type-annotation for executor

skrawcz:
See https://github.com/stitchfix/hamilton/pull/36 -- we changed the naming a little, and decided to punt on specific drivers for now. Until it's clear there needs to be more logic abstracted we're just sticking with the simple "GraphAdapters" for now.

So going to close this branch. Might resurrect this at a later time.

PR #36, opened 2022-01-30
Refactors distributed prototype

skrawcz:
The branch paves the way for distributed execution, and a general purpose framwork!

## Additions

1. Adds an interface to adapt execution of the hamilton function graph.
2. Adds prototype implementations to run hamilton in a distributed fashion: dask, ray, koalas on spark 3.2.
3. Adds the ability to not return a pandas dataframe. This opens up hamilton to a general purpose world!

## Removals

- old dask prototype

## Changes

- Adjusts graph.py and driver.py accordingly

## Testing

1. Adds tests for most new code.
2. Runs locally and on circleci.

## Screenshots
N/A

## Notes
N/A

## Todos
N/A

## Checklist

- [X] PR has an informative and human-readable title
~- [] Changes are limited to a single goal (no scope creep)~
~- [] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)~

### Python

- [x] python 3.6
- [x] python 3.7


hamilton/graph.py
```diff
@@ -97,16 +98,16 @@ def create_function_graph(*modules: ModuleType, config: Dict[str, Any]) -> Dict[
...
 
         :param modules: Modules to crawl for functions
-        :param config:
+        :param config: this is configuration and/or initial data.
+        :param executor: how we execute nodes in the graph
         """
         if executor is None:
-            from .driver import DirectExecutor
-            executor = DirectExecutor()
+            executor = base.SimplePythonDataFrameExecutor()
```
elijahbenizzy, line 110: Yeah, so this is making me think the `ResultsMixin` doesn't make sense. IMO it doesn't offer the flexibility out of the box for execution. Instead, we should have `ResultsGatherer` either: (a) be another parameter  to execute or (b) be a property of the driver, and add some structure into the driver. Embedding it as the default is not a decision we can easily roll back, either.

E.G. why would we want the default to return a dataframe, rather than return just a dict of variable names to values? We *do* want to make it really easy to return a dataframe, but we don't want that to require another class. 

Thinking about this in terms of responsibilities, we have the following:

1. FunctionGraph -- organizing the functions
2. ResultsBuilder -- putting it into a final stage. Perhaps this could even be another node in the graph?
3. Executor -- responsible for actually running everything

I think it makes sense to have the driver look like:

`return results_builder. build_result(function_graph.execute(executor=executor))`

And then have abstraction layers in the driver manage this.

hamilton/base.py
```diff
@@ -0,0 +1,105 @@
...
+    """Mixin for building a pandas dataframe from the result"""
+
+    @staticmethod
+    def build_result(**columns: typing.Dict[str, typing.Any]) -> pd.DataFrame:
+        # TODO check inputs are pd.Series, arrays, or scalars -- else error
+        # TODO do a basic index check across pd.Series?
+        return pd.DataFrame(columns)
+
+
+class NumpyMatrixResult(ResultMixin):
```
elijahbenizzy, line 42: This is cool. How is the column order specified?  I think this should take in an ordered dict or a list of columns + a list of results.

tests/test_graph.py
```diff
@@ -161,7 +161,7 @@ def create_testing_nodes():
 
 def test_execute():
     """Tests graph execution along with basic memoization since A is depended on by two functions."""
-    executor = driver.DirectExecutor()
+    executor = base.SimplePythonDataFrameExecutor()
```
elijahbenizzy, line 164: From a readability perspective how does this work? The `expected` is a dict but it outputs a dataframe.

hamilton/graph.py
skrawcz: Agreed. But we already have to contend with what's out in the wild -- so this is purely to maintain backwards compatibility. Furthermore, at this point in time, I don't see it being a problem to default to pandas.

elijahbenizzy: This *won't* be backwards compatible though? The current executor returns a dict, not a dataframe. The *driver* returns a dataframe.

hamilton/base.py
```diff
@@ -0,0 +1,105 @@
...
+                list_of_columns.append([val] * num_rows)
+            elif length == num_rows:
+                list_of_columns.append(list(val))
+            else:
+                raise ValueError(f'Do not know how to make this column {col} with length {length }have {num_rows} rows')
+        # Create the matrix with columns as rows and then transpose
+        return np.asmatrix(list_of_columns).T
+
+
+class HamiltonExecutor(ResultMixin):
```
elijahbenizzy, line 76: I think this is the wrong interface for the `Executor`. Rather, this is a node-based executor. See original design doc [here](https://docs.google.com/document/d/17LjBzv2hvSs8BSbjeKasWjY6UAMKxrE1En5bjkkFhto/edit#heading=h.y3wb8ql7gb9k)

This proposes two layers:
- An overall executor
- A topological sort executor that executes in serial.

Let's have this be a NodeWalkExecutor, that walks the graph and executes in order, abstracting away the `execute_node` component. Then we're simplifying `function_graph` by extracting the topological sort component into here (making it more unit testable), and enabling us to build other types of executors (E.G. passing the whole thing to spark for instance). Yes, more complexity, but I think this opens up a ton of doors.

hamilton/graph.py
elijahbenizzy: Yep, see current `execute_static` https://github.com/stitchfix/hamilton/blob/main/hamilton/graph.py#L272.
I think that if we bury assumptions like this into the system (pandas), we'll end up designing ourselves into a corner.

hamilton/base.py
skrawcz: I'm a yes on the last paragraph, but an unsure for everything above it. 

We might want to move away from the word Executor in general. I don't want to anchor to what people might think it is, given it's a term used in other contexts.

tests/test_graph.py
skrawcz: Because `build_result` isn't called in this call path :)

hamilton/base.py
skrawcz: Good thought -- haven't unit tested this function -- but yeah things to think about.

hamilton/graph.py
skrawcz: I'm confused, the executor in the context of the functionGraph is only called for type checking, and "executing" a node.

hamilton/base.py
elijahbenizzy: Sure, happy to think about other terminology here.

skrawcz: renamed things

hamilton/graph.py
skrawcz: I think we cleared this up IRL.

hamilton/base.py
```diff
@@ -0,0 +1,137 @@
...
+        :param input_type: The type of the input that would flow into the node.
+        :return:
+        """
+        pass
+
+    @abc.abstractmethod
+    def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) -> typing.Any:
+        """Given a node that represents a hamilton function, execute it.
+        Note, in some executors this might just return some type of "future".
+        :param node:
```
elijahbenizzy, line 118: Finish the docstrings?

hamilton/base.py
```diff
@@ -0,0 +1,137 @@
...
+            else:
+                raise ValueError(f'Do not know how to make this column {col} with length {length }have {num_rows} rows')
+        # Create the matrix with columns as rows and then transpose
+        return np.asmatrix(list_of_columns).T
+
+
+class HamiltonGraphAdapter(ResultMixin):
+    """Any GraphAdapters should implement this interface to adapt the HamiltonGraph for that particular context.
+
+    Note since it inherits ResultMixin -- HamiltonGraphAdapters need a `build_result` function.
```
elijahbenizzy, line 86: Still not sure it should -- rather the driver should use composition with the GraphAdapter and a resultsBuilder, but I think this is fine for now.

hamilton/driver.py
```diff
@@ -32,15 +32,17 @@ class Variable:
 class Driver(object):
     """This class orchestrates creating and executing the DAG to create a dataframe."""
 
-    def __init__(self, config: Dict[str, Any], *modules: ModuleType, executor=None):
+    def __init__(self, config: Dict[str, Any], *modules: ModuleType, executor: base.HamiltonGraphAdapter = None):
```
elijahbenizzy, line 35: Change parameter name

Issue #45, opened 2022-02-03
Specify path to save .gv files


latlan1:
## Current & Expected behavior
I wish there was a way of specifying the path for saving the DAG (.gv). I had two sets of DAGs and the file gets overwritten as test_output/execute.gv.

## Library & System Information
python version= 3.7.11, hamilton library version =1.1.1, linux= yes, Ubuntu 20 installed via WSL 1

skrawcz:
# Proposed API - split display out into a `visualize_execute` function
Reasoning -- much like the `explain` functionality in a DB -- we should expose an 'visualize_execute' function in the driver.
I don't want to call it `explain` because it's not obvious what to expect. I think `visualize_execute` is better.

In future if we add an `explain` functionality, it could delegate to this function for instance.

```python

class Driver(object):
    ...
    def visualize_execute(self, final_vars: List[str],
                overrides: Dict[str, Any] = None, output_file_uri: str,  **render_kwargs):
          """render_kwargs would be passed through to the graphviz render function"""
```

## deprecate display_graph in driver.execute
1 . We should deprecate this -- and tell people to move to the above function.
2. We should remove the cycle check that was thrown into the display function and move to it's own function.

skrawcz:
This has been merged to `main`. This will come out in the `1.3.0` release.

PR #58, opened 2022-02-06
Stefan/refactor visualization

skrawcz:
Visualization was an afterthought we added quickly.
This change takes in feedback from https://github.com/stitchfix/hamilton/issues/45 and makes
it more amenable for actual use.

We therefore expose new functions, and deprecate
the `visualize` parameter  in `driver.execute()`.

The new functions enable the user to pass in a dictionary that will
be used as key word arguments (kwargs) to the
render function -- thereby giving lots of control on
how things are visualized. We also mirror the changes
in display_all_functions().

Along the way I also refactored the cycle detection
using networkx. This again, was a quick hack,
now it's a standalone implementation. We do break graph
 behavior, but not driver behavior with this refactor. I
believe that's fine, since we don't want people to
depend on FunctionGraph directly just yet.

Adds unit tests for these changes.

## Additions

- specific helper functions to create graphviz and networkx graphs.
- new functions to display and enable passing

## Removals

- N/A

## Changes

- Deprecates driver.execute display parameter
- changes display_all_functions in a backwards compatible manner.

## Testing

1. Locally
2. via unit tests

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Code follows the standards laid out in the [TODO link to standards]()
- [ ] Placeholder code is flagged / future todos are captured in comments

### Python
Loosely tested across different python versions.
- [x] python 3.6
- [x] python 3.7
- [x] python 3.8
- [x] python 3.9

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -81,11 +81,14 @@ class Driver(object):
         """Executes computation.
 
         :param final_vars: the final list of variables we want in the data frame.
-        :param overrides: the user defined overrides.
-        :param display_graph: whether we want to display the graph being computed.
+        :param overrides: the user defined input variables.
+        :param display_graph: DEPRECATED. Whether we want to display the graph being computed.
         :param inputs: Runtime inputs to the DAG
         :return: a data frame consisting of the variables requested.
         """
+        if display_graph:
+            logger.warning('display_graph=True is deprecated. It will be removed in the 2.0.0 release. '
+                           'Please use visualize_execution().')
         columns = self.raw_execute(final_vars, overrides, display_graph, inputs=inputs)
         return self.adapter.build_result(**columns)
 
@@ -101,19 +104,16 @@ class Driver(object):
 
         :param final_vars: Final variables to compute
         :param overrides: Overrides to run.
+        :param display_graph: DEPRECATED. DO NOT USE. Whether or not to display the graph when running it
         :param inputs: Runtime inputs to the DAG
-        :param display_graph: Whether or not to display the graph when running it
         :return:
         """
         nodes, user_nodes = self.graph.get_required_functions(final_vars)
-
         self.validate_inputs(user_nodes, inputs)  # TODO -- validate within the function graph itself
         if display_graph:
-            # TODO: fix hardcoded path.
-            try:
-                self.graph.display(nodes, user_nodes, output_file_path='test-output/execute.gv')
-            except ImportError as e:
-                logger.warning(f'Unable to import {e}', exc_info=True)
+            self.visualize_execution(final_vars, 'test-output/execute.gv', {'view': True})
+            if self.has_cycles(final_vars):  # here for backwards compatible driver behavior.
+                raise ValueError('Error: cycles detected in you graph.')
         memoized_computation = dict()  # memoized storage
         self.graph.execute(nodes, memoized_computation, overrides, inputs)
         columns = {c: memoized_computation[c] for c in final_vars}  # only want request variables in df.
@@ -127,9 +127,46 @@ class Driver(object):
         """
         return [Variable(node.name, node.type) for node in self.graph.get_nodes()]
 
-    def display_all_functions(self):
-        """Displays the graph."""
-        self.graph.display_all()
+    def display_all_functions(self, output_file_uri: str):
+        """Displays the graph of all functions loaded!
+
+        :param output_file_uri: the full URI of path + file name to save the dot file to.
+            E.g. 'some/path/graph-all.dot'
+        """
+        try:
+            self.graph.display_all(output_file_uri)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def visualize_execution(self,
+                            final_vars: List[str],
+                            output_file_uri: str,
+                            render_kwargs: dict):
+        """Visualizes Execution.
+
+        Note: overrides are not handled at this time.
+
+        :param final_vars: the outputs we want to compute.
+        :param output_file_uri: the full URI of path + file name to save the dot file to.
+            E.g. 'some/path/graph.dot'
+        :param render_kwargs: a dictionary of values we'll pass to graphviz render function.
+        """
+        nodes, user_nodes = self.graph.get_required_functions(final_vars)
+        self.validate_inputs(user_nodes, self.graph.config)
+        try:
+            self.graph.display(nodes, user_nodes, output_file_uri, render_kwargs=render_kwargs)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def has_cycles(self, final_vars: List[str]) -> bool:
+        """Checks that the created graph does not have cycles.
+
+        :param final_vars: the outputs we want to comute.
+        :return: boolean True for cycles, False for no cycles.
+        """
+        nodes, user_nodes = self.graph.get_required_functions(final_vars)
+        self.validate_inputs(user_nodes, self.graph.config)
+        return self.graph.has_cycles(nodes, user_nodes)
 
 
 if __name__ == '__main__':
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -110,7 +110,9 @@ class Driver(object):
         """
         nodes, user_nodes = self.graph.get_required_functions(final_vars)
         self.validate_inputs(user_nodes, inputs)  # TODO -- validate within the function graph itself
-        if display_graph:
+        if display_graph:  # deprecated flow.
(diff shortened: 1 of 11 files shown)


hamilton/driver.py
```diff
@@ -76,9 +76,12 @@ def execute(self,
 
         :param final_vars: the final list of variables we want in the data frame.
         :param overrides: the user defined input variables.
-        :param display_graph: whether we want to display the graph being computed.
+        :param display_graph: DEPRECATED. Whether we want to display the graph being computed.
         :return: a data frame consisting of the variables requested.
         """
+        if display_graph:
```
elijahbenizzy, line 82: For a little later, let's create a `@deprecated` that does this for us.

```
@deprecated('display_graph`, message="display_graph=True is deprecated. It will be removed in the 2.0.0 release. Please use visualize_execution()", gone_at="2.0.0")
def execute(...):
    ...
```

hamilton/driver.py
```diff
@@ -117,9 +120,49 @@ def list_available_variables(self) -> List[Variable]:
         """
         return [Variable(node.name, node.type) for node in self.graph.get_nodes()]
 
-    def display_all_functions(self):
-        """Displays the graph."""
-        self.graph.display_all()
+    def display_all_functions(self, output_file_uri: str, render_kwargs: dict = None):
+        """Displays the graph of all functions loaded!
+
+        :param output_file_uri: the full URI of path + file name to save the dot file to.
```
elijahbenizzy, line 126: Filepath or URI (same thing but its worth thinking through)? Can we write to anything that's not a file (s3 location, EG?)

hamilton/driver.py
```diff
@@ -117,9 +120,49 @@ def list_available_variables(self) -> List[Variable]:
...
+        Note: overrides are not handled at this time.
+
+        :param final_vars: the outputs we want to compute.
+        :param output_file_uri: the full URI of path + file name to save the dot file to.
+            E.g. 'some/path/graph.dot'
+        :param render_kwargs: a dictionary of values we'll pass to graphviz render function. Defaults to viewing.
+            If you do not want to view the file, pass in `{'view':False}`.
+        """
+        nodes, user_nodes = self.graph.get_required_functions(final_vars)
+        self.validate_inputs(user_nodes, self.graph.config)
```
elijahbenizzy, line 151: This is going to break with my changes. And with runtime_inputs being required it's not going to work anyway.

hamilton/driver.py
```diff
@@ -117,9 +120,49 @@ def list_available_variables(self) -> List[Variable]:
...
+        try:
+            self.graph.display(nodes, user_nodes, output_file_uri, render_kwargs=render_kwargs)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def has_cycles(self, final_vars: List[str]) -> bool:
+        """Checks that the created grpah does not have cycles.
```
elijahbenizzy, line 158: spelling

hamilton/driver.py
```diff
@@ -117,9 +120,49 @@ def list_available_variables(self) -> List[Variable]:
...
+            If you do not want to view the file, pass in `{'view':False}`.
+        """
+        nodes, user_nodes = self.graph.get_required_functions(final_vars)
+        self.validate_inputs(user_nodes, self.graph.config)
+        try:
+            self.graph.display(nodes, user_nodes, output_file_uri, render_kwargs=render_kwargs)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def has_cycles(self, final_vars: List[str]) -> bool:
```
elijahbenizzy, line 157: If it has cycles we should be able to visualize it anyway? Might be a nice way to find the cycles?

hamilton/driver.py
```diff
@@ -117,9 +120,49 @@ def list_available_variables(self) -> List[Variable]:
...
+            E.g. 'some/path/graph-all.dot'
+        :param render_kwargs: a dictionary of values we'll pass to graphviz render function. Defaults to viewing.
+            If you do not want to view the file, pass in `{'view':False}`.
+        """
+        try:
+            self.graph.display_all(output_file_uri, render_kwargs)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def visualize_execution(self,
```
elijahbenizzy, line 136: We should really extract this into something pluggable. We could pass in a graphvisualizer that has this as the API, then call out to it. Or something like that. That said, we can always add later.

hamilton/driver.py
skrawcz: filepath to me is ambiguous -- but I suppose URI indicates a little more than we're capable of. hmm.

hamilton/driver.py
skrawcz: maybe. but not something we support right now.

hamilton/driver.py
skrawcz: yeah it needs to mirror it.

hamilton/driver.py
skrawcz: if we end up doing this a lot sure, else the extra code is probably not worth it :)

hamilton/driver.py
skrawcz: Yeah out of scope for now.

hamilton/driver.py
elijahbenizzy: What's ambiguous about filepath?

skrawcz: does it include the filename? or not

tests/test_graph.py
```diff
@@ -285,6 +287,96 @@ def test_config_can_override():
...
+    """Tests that display saves a file"""
+    fg = graph.FunctionGraph(tests.resources.dummy_functions, config={'b': 1, 'c': 2})
+    defined_nodes = set()
+    user_nodes = set()
+    for n in fg.get_nodes():
+        if n.user_defined:
+            user_nodes.add(n)
+        else:
+            defined_nodes.add(n)
+    expected = sorted(['// Dependency Graph\n',
```
elijahbenizzy, line 324: Why are we sorting this? Seems weird to compare a sorted file with what we expect?

hamilton/graph.py
```diff
@@ -160,16 +204,50 @@ def display_all(self, output_file_path: str = 'test-output/graph-all.gv'):
...
+        """Checks that the graph created does not contain cycles.
+
+        :param nodes: the set of nodes that need to be computed.
+        :param user_nodes: the set of inputs that the user provided.
+        :return: bool. True if cycles detected. False if not.
+        """
+        cycles = self.get_cycles(nodes, user_nodes)
+        return True if cycles else False
+
+    def get_cycles(self, nodes: Set[node.Node], user_nodes: Set[node.Node]) -> List:
```
elijahbenizzy, line 221: List of what? Not sure how we'd want to use this. Should be a list of paths (list of list of nodes?)

tests/test_graph.py
skrawcz: because the file does not have order guarantees.

skrawcz: it has the right contents, but not in the right order.

hamilton/graph.py
```diff
@@ -148,10 +190,12 @@ def config(self):
     def get_nodes(self) -> List[node.Node]:
         return list(self.nodes.values())
 
-    def display_all(self, output_file_path: str = 'test-output/graph-all.gv'):
-        """Displays the entire DAG structure constructed.
+    def display_all(self, output_file_path: str = 'test-output/graph-all.gv', render_kwargs: dict = None):
+        """Displays & saves a dot file of the entire DAG structure constructed.
```
elijahbenizzy, line 194: Might be nice to turn off the actual display piece so it can just generate

hamilton/driver.py
skrawcz: changing it to `output_file_path`

tests/test_graph.py
elijahbenizzy: Ahh a little weird but sure

skrawcz: e.g. it writes out a list of nodes, and a list of edges -- and how it outputs them isn't deterministic it turns out.

hamilton/graph.py
skrawcz: not the current behavior. :)

tests/test_graph.py
elijahbenizzy: Makes sense

hamilton/graph.py
skrawcz: added.

elijahbenizzy (approved):

PR #64, opened 2022-02-07
Lazy config evaluation

elijahbenizzy:

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Screenshots
If applicable

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Code follows the standards laid out in the [TODO link to standards]()
- [ ] Placeholder code is flagged / future todos are captured in comments

### Python

- [ ] python 3.6
- [ ] python 3.7


elijahbenizzy:
Need to add tests/figure out what to test. Hopefully this demonstrates the issue. IMO chainmap is actually better here -- supporting a lazyconfig is reasonable. That said, I'd prefer the simplicity with a dictionary.

elijahbenizzy:
Pros of this:
1. It allows lazily evaluated configs
2. Its idiomatic -- types are only as constricted as they need to be (E.G. why does config have to be a dict?)
3. It solves the problem a customer has

Cons of this:
1. We can do less with the config/inputs - E.G. if we want to store them somewhere or print them out we'll be running up against the same problem.
2. Added complexity for a specific use-case

hamilton/graph.py
```diff
@@ -310,10 +311,10 @@ def dfs_traverse(node: node.Node):
 
     @staticmethod
     def execute_static(nodes: Collection[node.Node],
-                       inputs: Dict[str, Any],
+                       inputs: Mapping[str, Any],
                        adapter: base.HamiltonGraphAdapter,
-                       computed: Dict[str, Any] = None,
-                       overrides: Dict[str, Any] = None):
+                       computed: Mapping[str, Any] = None,
+                       overrides: Mapping[str, Any] = None):
```
skrawcz, line 317: do these need to change?

hamilton/graph.py
```diff
@@ -379,7 +380,7 @@ def combine_config_and_inputs(config: Dict[str, Any], inputs: Dict[str, Any]) ->
         duplicated_inputs = [key for key in inputs if key in config]
         if len(duplicated_inputs) > 0:
             raise ValueError(f'The following inputs are present in both config and inputs. They must be mutually disjoint. {duplicated_inputs}')
-        return {**config, **inputs}
+        return collections.ChainMap(config, inputs)
```
skrawcz, line 383: does the unit test need to change to enforce chain map's way of not evaluating things?

elijahbenizzy: Yeah, if the contract is that we *don't* evaluate the driver then we want to test that its not evaluated. That said, I don't really like that as a contract, and it feels limiting later on, so let's think through.

hamilton/graph.py
elijahbenizzy: No, but I figured it would be nice to stay consistent.

PR #66, opened 2022-02-07
Adds simple case to help motivate @extract_fields

skrawcz:
If someone wanted to use Hamilton to model a modeling dataflow,
they would struggle. Need a new decorator to handle extracting
outputs from functions that return multiple things and aren't
a dataframe.

## Additions

- hello world showing how to create a model using hamilton as a dataflow language.
- adds decorator `@extract_fields`

## Testing

1. Unit tests & this hello world example.

## Checklist

- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### Python

- [x] python 3.6
- [x] python 3.7


examples/model_examples/scikit-learn/model_logic.py
```diff
@@ -0,0 +1,42 @@
...
+def flattened_digits(digit_data: np.ndarray) -> np.ndarray:
+    return digit_data.reshape((len(digit_data), -1))
+
+
+@function_modifiers.config.when(clf='svm')
+def prefit_clf__svm(gamma: float = 0.001) -> svm.SVC:
+    return svm.SVC(gamma=gamma)
+
+
+@function_modifiers.extract_columns(*['X_train', 'X_test', 'y_train', 'y_test'])
```
elijahbenizzy, line 20: Nit -- `*` is a little silly here :)

elijahbenizzy: The point is that this would currently break, right?

elijahbenizzy (commented):
Like it -- I think there are a lot of `extract_` we might want. Some options:

1. Have a single extract decorator that can assign polymorphically to different types that we can add extractable types easily
2. Have an extract decorator for each type, share some abstraction (perhaps (1)?)
3. Be opinionated about the things we want to extract -- only allow dfs, typeddicts, matrices (maybe...)

examples/model_examples/scikit-learn/model_logic.py
skrawcz: changed this.

skrawcz:
@elijahbenizzy I punted on TypedDict, in favor of something slightly simpler. 
We just need to add unit tests...

examples/model_examples/scikit-learn/run.py
```diff
@@ -0,0 +1,64 @@
...
+
+from hamilton import driver
+from hamilton import base
+
+import my_train_evaluate_logic
+import digit_loader
+import iris_loader
+
+
+def get_data_loader(data_set: str):
```
elijahbenizzy, line 13: Pros/cons of using this versus `@config.when` on the dataset? These were both recommended approaches I had in the bets practices btw.

examples/model_examples/scikit-learn/run.py
```diff
@@ -0,0 +1,64 @@
...
+    """Returns the module to load that will procur data -- the data loaders all have to define the same functions."""
+    if data_set == 'iris':
+        return iris_loader
+    elif data_set == 'digits':
+        return digit_loader
+    else:
+        raise ValueError(f'Unknown data_name {data_set}.')
+
+
+def get_model_config(model_type: str) -> dict:
```
elijahbenizzy, line 23: Might be nice to put these in json files -- cold be a simpler way to demonstrate storing/managing configs.

examples/model_examples/scikit-learn/run.py
elijahbenizzy: Yeah, thinking that this is probably best -- we should think through a way to codify the interfaces here.

hamilton/function_modifiers.py
```diff
@@ -190,6 +192,96 @@ def extractor_fn(column_to_extract: str = column, **kwargs) -> pd.Series:  # avo
         return output_nodes
 
 
+
+class extract_fields(NodeExpander):
+    """Extracts fields from a dictionary of output."""
+
+    def __init__(self, fields: dict, fill_with: Any = None):
```
elijahbenizzy, line 199: Fields should be `Dict[str, Type[Type]]` -- note that the keys have to be strings, and the values have to be keys.

hamilton/function_modifiers.py
```diff
@@ -190,6 +192,96 @@ def extractor_fn(column_to_extract: str = column, **kwargs) -> pd.Series:  # avo
...
+        - 1 function that outputs the original dict
+
+        :param fields: Fields to extract. A dict of 'field_name' -> 'field_type'.
+        :param fill_with: If you want to extract a field that doesn't exist, do you want to fill it with a default value?
+        Or do you want to error out? Leave empty/None to error out, set fill_value to dynamically create a field value.
+        """
+        if not fields:
+            raise InvalidDecoratorException('Error an empty dict, or no dict, passed to extract_fields decorator.')
+        elif not isinstance(fields, dict):
+            raise InvalidDecoratorException(f'Error, please pass in a dict or typed dict, not {type(fields)}')
```
elijahbenizzy, line 211: Why would fields be a typeddict? If so the types would be `Dict[str, Type[Type]]` or something like that. Again, see point above.

hamilton/function_modifiers.py
elijahbenizzy: If we choose to use a TypedDict, `fields` can be optional, validated, and type-safe. So a lot cleaner code but also trickier to read...

hamilton/function_modifiers.py
```diff
@@ -190,6 +192,96 @@ def extractor_fn(column_to_extract: str = column, **kwargs) -> pd.Series:  # avo
...
+        self.fields = fields
+        self.fill_with = fill_with
+
+    def validate(self, fn: Callable):
+        """A function is invalid if it is not annotated with a dict or typing.Dict return type.
+
+        :param fn: Function to validate.
+        :raises: InvalidDecoratorException If the function is not annotated with a dict or typing.Dict type as output.
+        """
+        output_type = inspect.signature(fn).return_annotation
```
elijahbenizzy, line 232: Let's be opinionated and force it to be a `Dict[str, Any]`. Again we can use the type-hint if its uniform...

hamilton/function_modifiers.py
```diff
@@ -190,6 +192,96 @@ def extractor_fn(column_to_extract: str = column, **kwargs) -> pd.Series:  # avo
...
+        if typing_inspect.is_generic_type(output_type):
+            base = typing_inspect.get_origin(output_type)
+            if base == dict or base == typing.Dict:
```
elijahbenizzy, line 235: Why would `base==dict` if its generic?

elijahbenizzy (commented):
Some thoughts 
We have three options for the API:
- Allow `dict`, force specification of types in decorator
- Allow `Dict[str, TYPE]`, force everything to be the same type, just specify names in decorator
- Allow TypedDict, specify everything in decorator

This chooses (1) -- let's think through the API a bit? IMO (2) is the most readable but its also limiting. Do we want to support dicts with varied value-types?

hamilton/function_modifiers.py
skrawcz: yep, hence punting on TypedDict support.

hamilton/function_modifiers.py
skrawcz: I have to change that string.

hamilton/function_modifiers.py
skrawcz: eh. I don't see the value of that code right now.

hamilton/function_modifiers.py
skrawcz: good point -- should be able to test for it and remove it if so -- I just forget what type it returns.

skrawcz: 3.6 returns `typing.Dict` 3.7+ returns dict.

elijahbenizzy (approved):

PR #83, opened 2022-02-24
Implements tags

elijahbenizzy:

## Additions

- Tagging features to nodes, variables

## Removals

- N/A

## Changes

- See above

## Testing

1. Added unit tests 
## Screenshots
If applicable

## Notes

-

## Todos

- [ ] Add docs

## Checklist

### Python

- [ ] python 3.6
- [x] python 3.7


elijahbenizzy:
Would also consider using keys/values as tags, but that seems more complex to manage.

skrawcz:
We should think how these tags fit into the overall picture of what we already capture.  
Spitballing, what if we wanted to capture what Lyft captures for features - how would we do it? E.g. https://eng.lyft.com/ml-feature-serving-infrastructure-at-lyft-d30bf2d3c32a has an example of the metadata they put on a feature. 

```json
"source": "hive", # this seems like a reasonable key-value.
"query_file": "blah.sql", # not relevant to us 
"pagerduty_emails": ["..."], # seems like this could be in a special decorator -- or done via tags.  Again a useful key-value.
"amundsen_tags": ["cancellation_xp"], # seems like amundsen doesn't use structure? In which case, is this what we want these tags to emulate?
"entity": "ride",  # this seems like a reasonable key-value.
"path": "...", # not sure what this is for...
"FEATURE_NAME": {
   "description": "foo bar",    # we should take this from the doc strings we already have.
   "type": "float",    # we should already have this (or think of some way to get it)
   "omit_values": [null, 0],   # in hamilton's world, this seems like specific transform data - -and shouldn't be metadata
   "has_personal_data": false,  # if this is boolean, then the presence of "has_personal_data" could suffice, rather than having a key-value for it.
}
```
So my intuition is that we will want to have some way to specify keys & values, but I guess it depends whether the tags are going to be a source for other functionality -- or functionality should explicitly request the information they need.

So I guess we need to get crisp on the use case for tags.

That said, I will add, a flat structure can have a hierarchy/key value structure, if we think in flattened terms, e.g. JSON path style. E.g. `entity.ride`, `has_personal_data.false`, `pagerduty_emails.foo@bar\.com,foo2@bar\.com` ...

elijahbenizzy:
> isp

Capturing some points from a conversation we had on use-cases

1. Recording whether a node has PII
2. Specifying components of a data product 
3. Specifying owners of a specific transform (team, etc...)

hamilton/function_modifiers.py
```diff
@@ -494,3 +494,52 @@ def resolves(configuration: Dict[str, Any]) -> bool:
...
+
+    @classmethod
+    def tag_allowed(cls, candidate: str) -> bool:
+        """Whether or not the candidate tag is allowed.
+        Its not allowed if it starts with a certain prefix.
+
+        :param candidate: Tag to check
+        :return: Whether or not its allowed
+        """
+        return not candidate.startswith(tag.RESERVED_TAG_PREFIX + '.')
```
skrawcz, line 532: we should make this check for things that aren't JSON friendly.

e.g. 
1. it's a string.
2. dots represent some way to namespace/segment.
3. special characters are not allowed (not sure what exactly this set is)
4.

elijahbenizzy: > we should make this check for things that aren't JSON friendly.
> 
> e.g.
> 
> 1. it's a string.
> 2. dots represent some way to namespace/segment.
> 3. special characters are not allowed (not sure what exactly this set is)
> 4. do we want some max length -- something unlikely to be hit, but a max nonetheless?

Yep, agreed -- although theoretically anything can be a key in JSON so not sure its necessary?

```
>>> import json
>>> json.loads('{"a[{" : "b}]"}')
{'a[{': 'b}]'}
```

That said, it might be nice anyway to limit it. If we assume everything is strings (which IMO is good), we should be fine with anything as long as `.` has a special meaning.

elijahbenizzy:
Now allows for hierarchy -- rules documented + tested. We'll also need some way of querying this in the driver -- will likely be adding soon. TBD on querying -- we'll make it so each tag set resolves to a json object and then we can query using a json object for a match.

skrawcz (approved):
I assume next PR would show/add functions to the driver?

elijahbenizzy:
> I assume next PR would show/add functions to the driver?

Yep -- left that open so I could ruminate over how to expose this.

elijahbenizzy:
Rewrote this a bit cause the original design felt off. Changes:
1) Made it a key/value API, using `**kwargs`
2) Made it so keys can be pathed strings (E.G. identifiers, separated by `.`). This will only be used by power-users, for now most will use the `**kwargs* API (although they can do `.`s with something like `@tag(**{"foo.bar" : "baz"})
3) Made it so the Variables API in `Driver` has a `tags` field that is a `dict[str, str]`

The `tags` field will have exactly what's inputted (not nested), but in the future I'll be adding tooling for more complex json queries.

hamilton/function_modifiers.py
```diff
@@ -494,3 +498,101 @@ def resolves(configuration: Dict[str, Any]) -> bool:
...
+            node_source=node_.node_source,
+            input_types=node_.input_types,
+            tags=self.tags)
+
+    @staticmethod
+    def _key_allowed(key: str) -> bool:
+        """Validates that a tag key is allowed. Rules are:
+        1. It must not be empty
+        2. It can have dots, which specify a hierarchy of order
+        3. All components, when split by dots, must be
```
skrawcz, line 550: incomplete sentence.

skrawcz (approved):
Looks good. Doing it via kwargs makes it look clean.

Question, do we want to augment doc strings on "functions" to also somehow expose tags? Just thinking of sphinx docs and how to pass tags through...

hamilton/function_modifiers.py
```diff
@@ -494,3 +498,101 @@ def resolves(configuration: Dict[str, Any]) -> bool:
...
+        """Validates the decorator. In this case that the set of tags produced is final.
+
+        :param fn: Function that the decorator is called on.
+        :raises ValueError: if the specified tags contains invalid ones
+        """
+        bad_tags = set()
+        for key, value in self.tags.items():
+            if (not tag._key_allowed(key)) or (not tag._value_allowed(value)):
+                import pdb
+                pdb.set_trace()
```
skrawcz, line 589: remove

PR #90, opened 2022-03-20
simplify pull request template

jameslamb:
This PR proposes some changes to `hamilton`'s pull request template. In my experience as a maintainer on other projects, I've found that the more information an issue / PR template asks for, the more likely it is that people will ignore it entirely.

Would you consider the following simplifications?

* condense "additions", "removals", and "changes" into just "changes"
* condense "todos", "screenshots", and "notes" into just "notes"

I think that this reduced set of sections still provides enough structure to encourage informative pull request descriptions.

Thanks very much for your time and consideration!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md 
+++ b/.github/PULL_REQUEST_TEMPLATE.md 
@@ -1,13 +1,5 @@
 [Short description explaining the high-level reason for the pull request]
 
-## Additions
-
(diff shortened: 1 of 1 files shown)


skrawcz (approved):
yep, thanks for the refinement. I think we can refine it/review it further after few more PRs :)

PR #95, opened 2022-03-23
Fixes has_cycles behavior

skrawcz:
We should memoize visiting the node before traversing dependencies.
This changes that behavior, changes the test, and checks that overrides
work and don't fail for a cyclic graph. That way we avoid a stack overflow error.

## Changes

- changing when a node is seen when determining the subgraph to execute.

## Testing

1. unit test work

## Checklist

- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### Python

- [x] python 3.6
- [x] python 3.7

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -298,8 +298,12 @@ def test_function_graph_has_cycles_true():
     nodes = [n for n in all_nodes if not n.user_defined]
     user_nodes = [n for n in all_nodes if n.user_defined]
     assert fg.has_cycles(nodes, user_nodes) is True
-    with pytest.raises(RecursionError):
-        fg.get_required_functions(['A', 'B', 'C'])
+    nodez, user_nodez = fg.get_required_functions(['A', 'B', 'C'])
+    assert nodez == set(nodes + user_nodes)
+    assert user_nodez == set(user_nodes)
+    result = fg.execute([n for n in nodes if n.name == 'B'], overrides={'A': 1, 'D': 2})
+    assert len(result) == 3
+    assert result['B'] == 3
 
 
 def test_function_graph_has_cycles_false():
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -292,18 +292,20 @@ def test_config_can_override():
 
 
 def test_function_graph_has_cycles_true():
-    """Tests whether we catch a graph with cycles"""
+    """Tests whether we catch a graph with cycles -- and expected behaviors"""
     fg = graph.FunctionGraph(tests.resources.cyclic_functions, config={'b': 2, 'c': 1})
     all_nodes = fg.get_nodes()
     nodes = [n for n in all_nodes if not n.user_defined]
     user_nodes = [n for n in all_nodes if n.user_defined]
     assert fg.has_cycles(nodes, user_nodes) is True
-    nodez, user_nodez = fg.get_required_functions(['A', 'B', 'C'])
-    assert nodez == set(nodes + user_nodes)
-    assert user_nodez == set(user_nodes)
+    required_nodes, required_user_nodes = fg.get_required_functions(['A', 'B', 'C'])
+    assert required_nodes == set(nodes + user_nodes)
+    assert required_user_nodes == set(user_nodes)
     result = fg.execute([n for n in nodes if n.name == 'B'], overrides={'A': 1, 'D': 2})
     assert len(result) == 3
     assert result['B'] == 3
+    with pytest.raises(RecursionError):  # throw recursion error when we don't have a way to short circuit
+        fg.execute([n for n in nodes if n.name == 'B'])
 
 
 def test_function_graph_has_cycles_false():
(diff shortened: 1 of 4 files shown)


elijahbenizzy:
I don't think this is a great idea. This means we can construct a DAG with cycles then execute it if we break them (with an override)? The clean _dag requires x inputs_ interface, all you have to do is give it_ will then be meaningless, as determining the inputs a DAG requires gets a lot more complicated. Furthermore, we'll create DAGs that aren't DAGs, meaning there's no guarentee about compilation ability to some other framework (E.G. spark). Or, rather, they're only compilable at runtime....

IIRC the use-case for this was symbolic equations (from feedback on reddit), but I don't think that's the use-case hamilton was built for.

tests/test_graph.py
```diff
@@ -296,8 +296,12 @@ def test_function_graph_has_cycles_true():
     nodes = [n for n in all_nodes if not n.user_defined]
     user_nodes = [n for n in all_nodes if n.user_defined]
     assert fg.has_cycles(nodes, user_nodes) is True
```
elijahbenizzy, line 298: `is True` should be changed to just `assert` btw (`is` is the object identity operator, which doesn't really work for this case (E.G. it doesn't evaluate truthiness -- == is what you want to be more explicit).

tests/test_graph.py
```diff
@@ -296,8 +296,12 @@ def test_function_graph_has_cycles_true():
...
-    with pytest.raises(RecursionError):
-        fg.get_required_functions(['A', 'B', 'C'])
+    nodez, user_nodez = fg.get_required_functions(['A', 'B', 'C'])
```
elijahbenizzy, line 299: What's the significanct of naming them with a `z`/why is `user_nodez` different from `user_nodes`?

skrawcz: to not collide with the variable defined above.

skrawcz:
> I don't think this is a great idea. This means we can construct a DAG with cycles then execute it if we break them (with an override)? The clean _dag requires x inputs_ interface, all you have to do is give it_ will then be meaningless, as determining the inputs a DAG requires gets a lot more complicated. Furthermore, we'll create DAGs that aren't DAGs, meaning there's no guarentee about compilation ability to some other framework (E.G. spark). Or, rather, they're only compilable at runtime....
> 
> IIRC the use-case for this was symbolic equations (from feedback on reddit), but I don't think that's the use-case hamilton was built for.

Here's why I think it's fine for us to support this:

1. We allow construction of a graph that in fact isn't a DAG -- we don't error during graph construction. This part just constructs our view of the world; we don't act on it yet. One could argue this doesn't have to be a DAG -- only at "execution" time does it need to be one.
2. We have inconsistent behavior with our functions -- and actually I'd argue it's a bug, one can't actually use the has_cycles() function in the toy example I came up with. So I want to clean that behavior up to correct it; the fix here is really to fix the DFS algo.
3. You're correct about compilation -- but we don't have that yet, and I don't think this work precludes any ability to be opinionated/implement that as we want.
4. This would be a "power user feature" -- and the code to support it is quite minimal and so I don't see much of a carrying cost;

elijahbenizzy:
> > I don't think this is a great idea. This means we can construct a DAG with cycles then execute it if we break them (with an override)? The clean _dag requires x inputs_ interface, all you have to do is give it_ will then be meaningless, as determining the inputs a DAG requires gets a lot more complicated. Furthermore, we'll create DAGs that aren't DAGs, meaning there's no guarentee about compilation ability to some other framework (E.G. spark). Or, rather, they're only compilable at runtime....
> > IIRC the use-case for this was symbolic equations (from feedback on reddit), but I don't think that's the use-case hamilton was built for.
> 
> Here's why I think it's fine for us to support this:
> 
> 1. We allow construction of a graph that in fact isn't a DAG -- we don't error during graph construction. This part just constructs our view of the world; we don't act on it yet. One could argue this doesn't have to be a DAG -- only at "execution" time does it need to be one.
> 2. We have inconsistent behavior with our functions -- and actually I'd argue it's a bug, one can't actually use the has_cycles() function in the toy example I came up with. So I want to clean that behavior up to correct it; the fix here is really to fix the DFS algo.
> 3. You're correct about compilation -- but we don't have that yet, and I don't think this work precludes any ability to be opinionated/implement that as we want.
> 4. This would be a "power user feature" -- and the code to support it is quite minimal and so I don't see much of a carrying cost;

Disagree but not passionately. Do we have a motivating use-case? The one on reddit was super interesting, but involved some pretty complex symbolic manipulation that hamilton didn't quite handle (but could!).

IMO if we support this explicitly it closes the door to a whole set of potential optimizations, and makes it more confusing to use. So, sure, easy to support, but IMO not worth it. E.G. other features will likely have to walk this back.

tests/test_graph.py
skrawcz: actually pycharm says `is` is better than `==` in this case...

tests/test_graph.py
skrawcz: fixed

tests/test_graph.py
skrawcz:
```
PEP 8: E712 comparison to True should be 'if cond is True:' or 'if cond:'
```

elijahbenizzy (approved):

PR #96, opened 2022-03-23
Enables full support for optional dependencies.

elijahbenizzy:
This was handled but not fully accounted for. The changes we made were:

1. Add compile-time checking of optionals (E.G. optionals don't have to exist)
2. Add runtime checking of optionals

To do (2) we had to change traversal of the function graph to account for
runtime inputs. This is because, in the driver, we ask for all functions
that are dependencies and then run them. We know what inputs we have,
so we can ask for all dependencies that we're going to need for a specific
execution. Thus we never ask for dependencies that don't need to exist

We should probably not always return the intermediate values, but we can
deal with that the next time we dig into materialization.

## Changes

-

## Testing

1.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -10,6 +10,8 @@ import typing
 
 from dataclasses import dataclass
 
+from hamilton import node
+
 if __name__ == '__main__':
     import graph
     import node
@@ -47,6 +49,22 @@ class Driver(object):
         self.graph = graph.FunctionGraph(*modules, config=config, adapter=adapter)
         self.adapter = adapter
 
+    def _node_is_required_by_anything(self, node_: node.Node) -> bool:
+        """Checks dependencies on this node and determines if at least one requires it.
+
+        Nodes can be optionally depended upon, i.e. the function parameter has a default value. We want to check that
+        of the nodes the depend on this one, at least one of them requires it, i.e. the parameter is not optional.
+
+        :param node_: node in question
+        :return: True if it is required by any downstream node, false otherwise
+        """
+        required = False
+        for downstream_node in node_.depended_on_by:
+            _, dep_type = downstream_node.input_types[node_.name]
+            if dep_type == node.DependencyType.REQUIRED:
+                return True
+        return required
+
     def validate_inputs(self, user_nodes: Collection[node.Node], inputs: typing.Optional[Dict[str, Any]] = None):
         """Validates that inputs meet our expectations. This means that:
         1. The runtime inputs don't clash with the graph's config
@@ -61,8 +79,9 @@ class Driver(object):
         errors = []
         for user_node in user_nodes:
             if user_node.name not in all_inputs:
-                errors.append(f'Error: Required input {user_node.name} not provided '
-                              f'for nodes: {[node.name for node in user_node.depended_on_by]}.')
+                if self._node_is_required_by_anything(user_node):
+                    errors.append(f'Error: Required input {user_node.name} not provided '
+                                  f'for nodes: {[node.name for node in user_node.depended_on_by]}.')
             elif (all_inputs[user_node.name] is not None
                   and not self.adapter.check_input_type(user_node.type, all_inputs[user_node.name])):
                 errors.append(f'Error: Type requirement mismatch. Expected {user_node.name}:{user_node.type} '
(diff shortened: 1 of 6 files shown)


elijahbenizzy:
Still a WIP. Need to figure out merging with the latest fixes...
- [ ] Fix the cycles test (not sure why its failing)
- [ ] Add documentation on optionals
- [ ] Add some optional-specific-tests

hamilton/node.py
```diff
@@ -63,21 +63,26 @@ def __init__(self,
...
                         self._input_types = {key: (value, DependencyType.REQUIRED) for key, value in input_types.items()}
             else:
                 signature = inspect.signature(callabl)
-                self._input_types = {}
                 for key, value in signature.parameters.items():
                     if value.annotation == inspect._empty:
                         raise ValueError(f'Missing type hint for {key} in function {name}. Please add one to fix.')
                     self._input_types[key] = (value.annotation, DependencyType.from_parameter(value))
+        elif self.user_defined:
+            if self.user_defined:
```
skrawcz, line 82: same clause in both?

hamilton/graph.py
```diff
@@ -322,17 +323,40 @@ def get_impacted_nodes(self, var_changes: List[str]) -> Set[node.Node]:
...
+        :param runtime_inputs: runtime inputs to the DAG -- if not provided we will assume we're running at compile-time.
+        Everything would then be required (even though it might be marked as optional), as we want to crawl the whole DAG.
+        If we're at runtime, we want to just use the nodes that are provided, and not count the optional ones that are not.
         :return: a tuple of sets:
             - set of all nodes.
             - subset of nodes that human input is required for.
         """
-        return self.directional_dfs_traverse(lambda n: n.dependencies, starting_nodes=final_vars)
+        def next_nodes_function(n: node.Node) -> List[node.Node]:
+            if runtime_inputs is None:
```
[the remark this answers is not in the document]
elijahbenizzy: I think we do

hamilton/driver.py
```diff
@@ -10,6 +10,8 @@
 
 from dataclasses import dataclass
 
+from hamilton.node import DependencyType
```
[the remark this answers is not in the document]
skrawcz: true

hamilton/driver.py
```diff
@@ -152,13 +168,13 @@ def visualize_execution(self,
         Note: overrides are not handled at this time.
 
         :param final_vars: the outputs we want to compute.
-        :param output_file_path: the full URI of path + file name to save the dot file to.
+        :param output_file_path: thefrom  full URI of path + file name to save the dot file to.
```
skrawcz, line 171:
```suggestion
        :param output_file_path: the  full URI of path + file name to save the dot file to.
```

elijahbenizzy:
Will add documentation after merging/publishing. @skrawcz mind an approval?

hamilton/driver.py
```diff
@@ -47,6 +49,19 @@ def __init__(self, config: Dict[str, Any], *modules: ModuleType, adapter: base.H
         self.graph = graph.FunctionGraph(*modules, config=config, adapter=adapter)
         self.adapter = adapter
 
+    def _node_is_required_by_anything(self, node_: node.Node) -> bool:
+        """Tells us if a node is required by any other node
+
```
skrawcz, line 54: Required here is a bit of a confusing term, i.e. if there's a link it's required in some sense. 

This is probably better to describe this funciton:
```suggestion
        """Checks dependencies on this node and determines if at least one requires it.
        
        Nodes can be optional, i.e. the function parameter has a default value, we want to check that
        of the nodes the depend on this one, at least one of them requires it, i.e. the parameter is not optional.
        
```

hamilton/driver.py
```diff
@@ -47,6 +49,19 @@ def __init__(self, config: Dict[str, Any], *modules: ModuleType, adapter: base.H
...
+        :param node_: node in question
+        :return: True if it is required by any downstream node, false otherwise
+        """
+        required = False
+        for downstream_node in node_.depended_on_by:
+            _, dep_type = downstream_node.input_types[node_.name]
+            if dep_type == node.DependencyType.REQUIRED:
+                required = True
```
skrawcz, line 62: optimization, can just short circuit here and return True.

skrawcz (approved):

PR #103, opened 2022-03-27
remove unused loggers

jameslamb:
Looking through this project's code today, I found several places where module-specific loggers are instantiated but never used.

This PR proposes removing them

## Changes

- removes import of `logging` and instantiation of unused loggers in docs, examples, and the package itself

## Testing

Ran the following to emulate CircleCI locally.

```shell
docker run \
   --rm \
    -v $(pwd):/usr/local/src \
    --workdir /usr/local/src \
   -it circleci/python:3.7 \
        /bin/bash
```

And then installed `hamilton` and ran its tests.

```shell
# install dependencies
sudo apt install graphviz
python -m venv venv || virtualenv venv
. venv/bin/activate
python --version
pip --version
pip install -r requirements-test.txt
pip install -r requirements.txt
pip install "dask[complete]"

# run tests
python -m pytest --cov=hamilton tests/
python -m pytest graph_adapter_tests/h_dask
```

## Notes

- removing unnecessary code from examples simplifies those examples and makes them more easier for new users to understand
- removing unnecessary imports and module-level instantiations reduces the time it takes to run `import hamilton`
- removing unused code reduces the time it takes to run static analyzers like `mypy` and `flake8`

Thanks for your time and consideration!

## Checklist

### Python - local testing

- [ ] python 3.6
- [x] python 3.7

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,10 @@ Note: functions can take or create scalar values, in addition to any python obje
 2. Create a `my_script.py` which is where code will live to tell Hamilton what to do:
 ```python
 import importlib
-import logging
-import sys
 
 import pandas as pd
 from hamilton import driver
 
-logger = logging.getLogger(__name__)
-logging.basicConfig(stream=sys.stdout)
 initial_columns = {  # load from actuals or wherever -- this is our initial data we use as input.
     # Note: these do not have to be all series, they could be scalar inputs.
     'signups': pd.Series([1, 10, 50, 100, 200, 400]),
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ import pandas as pd
 from hamilton import driver
 
 logging.basicConfig(stream=sys.stdout)
-
 initial_columns = {  # load from actuals or wherever -- this is our initial data we use as input.
     # Note: these do not have to be all series, they could be scalar inputs.
     'signups': pd.Series([1, 10, 50, 100, 200, 400]),
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ import importlib
 import pandas as pd
 from hamilton import driver
 
+logging.basicConfig(stream=sys.stdout)
+
 initial_columns = {  # load from actuals or wherever -- this is our initial data we use as input.
     # Note: these do not have to be all series, they could be scalar inputs.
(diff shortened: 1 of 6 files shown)


elijahbenizzy (approved):
Nice! Can't believe we had so much cruft...

README.md
```diff
@@ -76,14 +76,10 @@ Note: functions can take or create scalar values, in addition to any python obje
...
 ```python
 import importlib
-import logging
-import sys
 
 import pandas as pd
 from hamilton import driver
 
-logger = logging.getLogger(__name__)
-logging.basicConfig(stream=sys.stdout)
```
skrawcz, line 86: I believe I included this because I've seen people not know how to setup logging.  So I'm inclined to keep this line, but yeah, we don't need the logger  because we don't use it.

examples/hello_world/my_script.py
```diff
@@ -1,12 +1,8 @@
 import importlib
-import logging
-import sys
 
 import pandas as pd
 from hamilton import driver
 
-logger = logging.getLogger(__name__)
-logging.basicConfig(stream=sys.stdout)
```
skrawcz, line 9: ditto -- let's keep this.

skrawcz (changes requested):
I think we want to keep the logging setup lines. Since they control logging and are the entry point to running things. Otherwise yeah, if we're not using the loggers we should remove them -- can always add them back when we have something to log.

README.md
jameslamb: sure, can do! Restored in https://github.com/stitchfix/hamilton/pull/103/commits/85a1a82e81e3bcbdc12d9b127d8f61e1d022f292

examples/hello_world/my_script.py
jameslamb: ok no prob, restored in https://github.com/stitchfix/hamilton/pull/103/commits/85a1a82e81e3bcbdc12d9b127d8f61e1d022f292

skrawcz (approved):

skrawcz:
@jameslamb apologies for the slowness -- I miss github notifications often :/

jameslamb:
ha no problem!

PR #104, opened 2022-04-06
Adds parameterized_inputs

skrawcz:
This change adds `paramterized_inputs` decorator, which
is almost a carbon copy of `paramterized_input` but that it allows
any number of parameters to be mapped.

Why is it a separate class? Well for backwards compatibility
we don't want to break parameterized_input. Should we try to
consolidate them? I think so -- so we should then mark `parameterized_input`
as deprecated and will be removed in a 2.0 release? ✅ 

I should then probably update all documentation to reflect `parameterized_inputs`
and thus the documentation on `paramterized_input` to either be hidden or
non-existant? Hmm. ✅ 

## Changes

- Add parameterized_inputs decorator.
- Deprecates parameterized_input decorator.
- Adjust documentation to promote parameterized_inputs instead.

## Testing

1. unit tests pass.

## Notes
Motivation is a side project and having to write out repetitive functions is laborious -- this is admittedly a power
user mode, but one I think that makes sense to have.
e.g. this looks nice and readable
```python
from hamilton.function_modifiers import parametrized_inputs
import pandas as pd

NUMERIC_FEATURES = ['age', 'height', 'weight', 'body_mass_index', 'transportation_expense',
                    'distance_from_residence_to_work', 'service_time', 'work_load_average_per_day',
                    'hit_target']


@parametrized_inputs(**{f'{k}_mean': dict(feature=k) for k in NUMERIC_FEATURES})
def mean_computer(feature: pd.Series) -> pd.Series:
    """Average of {feature}"""
    return feature.mean()


@parametrized_inputs(**{f'{k}_std_dev': dict(feature=k) for k in NUMERIC_FEATURES})
def std_dev_computer(feature: pd.Series) -> pd.Series:
    """Standard deviation of {feature}"""
    return feature.std()


@parametrized_inputs(**{f'{k}_zero_mean': dict(feature=k, feature_mean=f'{k}_mean') for k in NUMERIC_FEATURES})
def zero_mean_computer(feature: pd.Series, feature_mean: pd.Series) -> pd.Series:
    """Creates zero mean value of {feature} by subtracting {feature_mean}"""
    return feature - feature_mean


@parametrized_inputs(**{f'{k}_zero_mean_unit_var': dict(feature_mean=f'{k}_mean', feature_std_dev=f'{k}_std_dev') 
                        for k in NUMERIC_FEATURES})
def zero_mean_unit_var_computer(feature_mean: pd.Series, feature_std_dev: pd.Series) -> pd.Series:
    """Computes zero mean unit variance of {feature_mean} with {feature_std_dev} to create {output_name}"""
    return feature_mean / feature_std_dev

```

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7

### DIFF
--- a/decorators.md
+++ b/decorators.md
@@ -29,42 +29,46 @@ distinct outputs. The _parameter_ key word argument has to match one of the argu
 the arguments are pulled from outside the DAG. The _assigned_output_ key word argument takes in a dictionary of
 tuple(Output Name, Documentation string) -> value.
 
-## @parametrized_input
-Expands a single function into n, each of which corresponds to a function in which the parameter value is fed
-the input from a *specific column*. Note this decorator and `@parametrized` are quite similar, except that
-the input here is another DAG node, i.e. column, rather than some specific value.
+## @parametrized_inputs
+Expands a single function into _n_, each of which corresponds to a function in which the parameters specified are mapped
+to the specified inputs. Note this decorator and `@parametrized` are quite similar, except that
+the input here is another DAG node(s), i.e. column/input, rather than a specific scalar/static value.
 ```python
 import pandas as pd
-from hamilton.function_modifiers import parametrized_input
-import internal_package_with_logic
+from hamilton.function_modifiers import parametrized_inputs
 
-ONE_OFF_DATES = {
-     #input var        (# output var,               # description of new outputs)
-     'D_ELECTION_2016', ('D_ELECTION_2016_shifted', 'US election 2016 shifted by 1'),
-     'SOME_INPUT_NAME', ('SOME_OUTPUT_NAME', 'Doc string for this thing'),
-}
-            # parameter matches the name of the argument in the function below
-@parametrized_input(parameter='one_off_date', assigned_inputs=ONE_OFF_DATES)
+
+@parametrized_inputs(
+    D_ELECTION_2016_shifted=dict(one_off_date='D_ELECTION_2016'),
+    SOME_OUTPUT_NAME=dict(one_off_date='SOME_INPUT_NAME')
+)
 def date_shifter(one_off_date: pd.Series) -> pd.Series:
+    """{one_off_date} shifted by 1 to create {output_name}"""
     return one_off_date.shift(1)
 
 ```
-We see here that `parameterized_input` allows you to keep your code DRY by reusing the same function to create multiple
-distinct outputs. The _parameter_ key word argument has to match one of the arguments in the function. The rest of
-the arguments are pulled from items inside the DAG. The _assigned_inputs_ key word argument takes in a
-dictionary of input_column -> tuple(Output Name, Documentation string).
+We see here that `parameterized_inputs` allows you to keep your code DRY by reusing the same function to create multiple
+distinct outputs. The key word arguments passed have to have the following structure:
+> OUTPUT_NAME = Mapping of function argument to input that should go into it.
 
-Note that this is equivalent to writing the following two function definitions:
+So in the example, `D_ELECTION_2016_shifted` is an _output_ that will correspond to replacing `one_off_date` with `D_ELECTION_2016`.
+Then similarly `SOME_OUTPUT_NAME` is an _output_ that will correspond to replacing `one_off_date` with `SOME_INPUT_NAME`.
+The documentation for both uses the same function doc and will replace values that are templatized with the input
+parameter names, and the reserved value `output_name`.
+
+To help visualize what the above is doing, it is equivalent to writing the following two function definitions:
 
 ```python
 def D_ELECTION_2016_shifted(D_ELECTION_2016: pd.Series) -> pd.Series:
+    """D_ELECTION_2016 shifted by 1 to create D_ELECTION_2016_shifted"""
     return D_ELECTION_2016.shift(1)
 
 def SOME_OUTPUT_NAME(SOME_INPUT_NAME: pd.Series) -> pd.Series:
+    """SOME_INPUT_NAME shifted by 1 to create SOME_OUTPUT_NAME"""
     return SOME_INPUT_NAME.shift(1)
 ```
 
-Note also that the different input variables must all have compatible types with the original decorated input variable.
+*Note*: that the different input variables must all have compatible types with the original decorated input variable.
 
 ## @extract_columns
 This works on a function that outputs a dataframe, that we want to extract the columns from and make them individually
--- a/hamilton/function_modifiers.py
+++ b/hamilton/function_modifiers.py
@@ -1,7 +1,8 @@
 import functools
+import logging
 import inspect
 import typing
-from typing import Dict, Callable, Collection, Tuple, Union, Any, Type
+from typing import Dict, Callable, Collection, Tuple, Union, Any, Type, List, NamedTuple
 import typing_inspect
 
 import pandas as pd
@@ -9,6 +10,9 @@ import pandas as pd
 from hamilton import node
 from hamilton.function_modifiers_base import NodeCreator, NodeResolver, NodeExpander, sanitize_function_name
 from hamilton.models import BaseModel
+
+
+logger = logging.getLogger(__name__)
 
 """
 Annotations for modifying the way functions get added to the DAG.
@@ -72,7 +76,7 @@ class parametrized_input(NodeExpander):
         Note this decorator and `@parametrized` are quite similar, except that the input here is another DAG node,
         i.e. column, rather than some specific value.
 
-        The `parameterized_inputs` allows you keep your code DRY by reusing the same function but replace the inputs
+        The `parameterized_input` allows you keep your code DRY by reusing the same function but replace the inputs
         to create multiple corresponding distinct outputs. The _parameter_ key word argument has to match one of the
         arguments in the function. The rest of the arguments are pulled from items inside the DAG.
         The _assigned_inputs_ key word argument takes in a
@@ -97,6 +101,7 @@ class parametrized_input(NodeExpander):
             specific_inputs[input_column] = specific_inputs.pop(self.parameter)  # replace the name with the new function name so we get the right dependencies
 
             def new_fn(*args, input_column=input_column, **kwargs):
+                """This function rewrites what is passed in kwargs to the right kwarg for the function."""
                 kwargs = kwargs.copy()
                 kwargs[self.parameter] = kwargs.pop(input_column)
                 return node_.callable(*args, **kwargs)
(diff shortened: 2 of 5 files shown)


skrawcz:
So next steps I'm thinking:

1. Change docs to default to `parameterized_inputs` (plural).
2. Put a deprecation warning on `parameterized_input` -- and log when it's used.

tests/resources/parametrized_inputs.py
```diff
@@ -18,3 +22,28 @@ def input_2() -> int:
 )
 def function_with_multiple_inputs(input_value_tbd: int, static_value: int) -> int:
     return input_value_tbd + static_value
+
+
+@parametrized_inputs(
+    'input_value_tbd1', 'input_value_tbd2',
+    variable_inputs={
+        ('input_1', 'input_2'): ('output_12', 'function_with_multiple_inputs called using input_1'),
```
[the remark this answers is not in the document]
skrawcz: perhaps a namedtuple/object would be better? hmm

hamilton/function_modifiers.py
```diff
@@ -122,6 +127,73 @@ def validate(self, fn: Callable):
...
+                    f'assigned_output key is incorrect: {node}. The parameterized decorator needs a dict of '
+                    f'tuples to tuples: (input column1, ..., input column N) -> (name, description) to function.')
+
+    def expand_node(self, node_: node.Node, config: Dict[str, Any], fn: Callable) -> Collection[node.Node]:
+        nodes = []
+        input_types = node_.input_types
+        for input_column, (node_name, node_description) in self.assigned_output.items():
+            specific_inputs = {key: value for key, (value, _) in input_types.items()}
+            for idx, col in enumerate(input_column):
+                logger.info(f'For function {node_.name}: mapping {col} to {self.parameters[idx]}.')
```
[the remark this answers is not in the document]
skrawcz: To get at the "what is this doing" comment above, and alleviate possible sources of error, I'm thinking `info`. This would only be called on DAG creation so I think `info` is fine.

tests/resources/parametrized_inputs.py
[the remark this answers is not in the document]
skrawcz: or I just make the first argument be a list, rather than a variable length thing... hmm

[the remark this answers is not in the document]
skrawcz: see latest commit -- I think named tuple makes it look much nicer. Also renamed the arguments.

hamilton/function_modifiers.py
```diff
@@ -122,6 +127,73 @@ def validate(self, fn: Callable):
                 f'Annotation is invalid -- no such parameter {self.parameter} in function {fn}')
 
 
+class parametrized_inputs(NodeExpander):
+
+    def __init__(self, *parameters: str, variable_inputs: Dict[Tuple, Tuple[str, str]]):
+        """Constructor for a modifier that expands a single function into n, each of which
+        corresponds to the specified parameter replaced by a *specific input column*.
```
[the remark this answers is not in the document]
skrawcz: done.

elijahbenizzy:
Not completely sure how I feel about this API. Definitely cleaner but it seems that its not particularly consistent with other APIs. Its dryer/slightly easier to read (which is good), but it takes quite a bit of work to figure out what's happening/parse the API. The design philosophy behind decorators is specifically to make it easy to figure out what's going on. I worry that with objects/named tuples its going to be harder to figure out the shape of the DAG.

Honestly, I wonder if that's just an inherent issue with `@parametrized_inputs`?

decorators.md
```diff
@@ -29,30 +29,33 @@ distinct outputs. The _parameter_ key word argument has to match one of the argu
...
+ONE_OFF_DATES = [
+    ParameterMapping(inputs=['D_ELECTION_2016'],
+                     output_name='D_ELECTION_2016_shifted',
+                     output_docs='US election 2016 shifted by 1'),
+    ParameterMapping(inputs=['SOME_INPUT_NAME'],
+                     output_name='SOME_OUTPUT_NAME',
+                     output_docs='Doc string for this thing'),
+]
+
+@parametrized_inputs(parameters=['one_off_date'], parameter_mappings=ONE_OFF_DATES)
```
elijahbenizzy, line 49: Let's not have constants here IMO. Makes it slightly more confusing (I know its from the previous one, but still).

decorators.md
```diff
@@ -29,30 +29,33 @@ distinct outputs. The _parameter_ key word argument has to match one of the argu
...
-     #input var        (# output var,               # description of new outputs)
-     'D_ELECTION_2016', ('D_ELECTION_2016_shifted', 'US election 2016 shifted by 1'),
-     'SOME_INPUT_NAME', ('SOME_OUTPUT_NAME', 'Doc string for this thing'),
-}
-            # parameter matches the name of the argument in the function below
-@parametrized_input(parameter='one_off_date', assigned_inputs=ONE_OFF_DATES)
+from hamilton.function_modifiers import parametrized_inputs, ParameterMapping
+
+ONE_OFF_DATES = [
+    ParameterMapping(inputs=['D_ELECTION_2016'],
```
elijahbenizzy, line 41: Inputs as a list feels sketchy to me. Ieverything in the framework is `**kwargs`, and the mapping is what we want to capture with this API. Currently the mapping you have is:
(1) A list of `inputs` in `ParameterMapping` to (2) a corresponding list of parameters.  This requires some jumping around in reading it -- the natural shape is an actual map. 

There's also the interesting edge case of an input from `@parametrized_inputs` overriding a node that's already in the DAG. In that case, you may want to parametrize one *and* two different inputs. E.G.


```python
# I might want to run this with: [bar=bar, baz=baz]
# Or with [bar=bar, baz=foobaz]
# Or with [bar=foobar, baz=foobaz]
def foo(bar, baz):
    ...
```
Not sure if this is legit or not, honestly. However, with default arguments, you could have a need for differently shaped inputs assignment...

elijahbenizzy (commented):
Some thoughts

decorators.md
skrawcz: in reality that's what one would do for legibility. So I think this is fine, even though a little contrived for this example? I can try it on and see before making up my mind.

decorators.md
skrawcz: The issue is documentation.

This gets at your point -- but where do you stick documentation?
```python
@parameterized_inputs(
   {
     output1=dict(arg1=node_foo, arg2=node_bar), 
     output2=dict(arg1=node_bar, arg2=node_zzz)
   }
)
def zero_mean(arg1: type, arg2: type) -> type:
```

Oh wait maybe this could work:
```python
@parameterized_inputs(
   {
     output1=dict(arg1=node_foo, arg2=node_bar), 
     output2=dict(arg1=node_bar, arg2=node_zzz)
   }
)
def zero_mean(arg1: type, arg2: type) -> type:
     """Creates {output} which is the zero mean value of {arg1}. 
     
     :param {arg1}: {arg1},  the original value
     :param {arg2}: {arg2}, the mean value
     :return: a zero mean value..
     """
     return arg1 - arg2
```
🤔  will tinker some more here.

elijahbenizzy: Yep as I said in the DM I like the second option. Also encourages writing of documentation only once.

decorators.md
skrawcz: done.

decorators.md
skrawcz: done.

hamilton/function_modifiers.py
```diff
@@ -122,6 +129,85 @@ def validate(self, fn: Callable):
...
+            nodes.append(
+                node.Node(
+                    node_name,
+                    node_.type,
+                    node_description,
+                    new_fn,
+                    input_types=specific_inputs))
+        return nodes
+
+    def validate(self, fn: Callable):
```
elijahbenizzy, line 189: Can we validate documentation as well (I think we're not doing this...)

hamilton/function_modifiers.py
```diff
@@ -122,6 +129,85 @@ def validate(self, fn: Callable):
                 f'Annotation is invalid -- no such parameter {self.parameter} in function {fn}')
 
 
+class parametrized_inputs(NodeExpander):
+    RESERVED_KWARG = 'output_name'
+
+    def __init__(self, parameterization: Dict[str, Dict[str, str]]):
```
elijahbenizzy, line 135: Why not use `**parameterization` instead of the single argument? You can (a) still have reserved kwargs, and (b) can use the `**` to send in non-python-variable parametrizations if you want (but we don't).

elijahbenizzy (commented):
LGTM -- not convinced we want `parameterization`. I get it if we need later kwargs, but I think using `**parameterization` is going to look cleaner/be more readable.

hamilton/function_modifiers.py
skrawcz: I have it in the tests.

It will error if a template variable is not passed; otherwise up to the user to put the templates variables in.

hamilton/function_modifiers.py
skrawcz: oh hmm.

Not sure I like that, e.g. 

```python
def bar(**k):
    print(k)

>> bar(q={'a':1})
{'q': {'a': 1}}

>> bar(**{'a':1})
{'a': 1}
```

To me it's a little confusing what the structure of the dictionary should be from reading the code. 

>  You can (a) still have reserved kwargs

To me this means:
```python
def bar(*, arg1, arg2, **kwargs):
```
But I don't think we need to stick it in kwargs; to me that's less legible? i.e. the second person coming to read some code someone already wrote is less readable.


> (b) can use the ** to send in non-python-variable parametrizations if you want (but we don't).

I don't think so. We wouldn't know how to differentiate between a string value and the name of an input? I did consider whether this could consolidate `parameterized` but because it's hard to know what's a scalar value or the name of an input, I don't think it's a good idea... or would require cluttering the API. But then again, shouldn't decorator layering help here?

elijahbenizzy: Not what I'm talking about I think?


With `parameterization`:
```python
@parametrized_inputs(
    parameterization=dict(output_12={'input_value_tbd1': 'input_1', 'input_value_tbd2': 'input_2'})
)
def function_with_two_parameters(input_value_tbd1: int,
                                 input_value_tbd2: int,
                                 static_value: int) -> int:
```

With kwargs

```python
@parametrized_inputs(
    output_12={'input_value_tbd1': 'input_1', 'input_value_tbd2': 'input_2'},
)
def function_with_two_parameters(input_value_tbd1: int,
                                 input_value_tbd2: int,
                                 static_value: int) -> int:
```

I think the second is a lot cleaner. Arguments against it are:

1. What if you don't have a valid python variable (you should, but it works with `**{}`)
2. What if you want to add more kwargs to `@parameterized` (we still can, but its messy potentially)

Arguments for:

1. Its cleaner -- it represents a clear assignment of outputs to input mappings
2. Its less verbose
3. Its consistent with others (say `config.when`) that use kwargs

hamilton/function_modifiers.py
```diff
@@ -122,6 +129,85 @@ def validate(self, fn: Callable):
...
+        :param fn: Function to validate against this annotation.
+        :raises: InvalidDecoratorException If the function does not have the requested parameter
+        """
+        signature = inspect.signature(fn)
+        func_param_name_set = set(signature.parameters.keys())
+        if self.RESERVED_KWARG in func_param_name_set:
+            raise InvalidDecoratorException(
+                f'Error function {fn.__module__}.{fn.__name__} cannot have `{self.RESERVED_KWARG}` '
+                f'as a parameter it is reserved.')
+        errors = set()
```
elijahbenizzy, line 201: nit: this should be called `missing_parameters`

hamilton/function_modifiers.py
elijahbenizzy: Maybe I'm missing it but we don't test that/how it errors when the documentation has a misspelled parameter or something. Given how this is a little magical (E.G. uses {} + python's formatter), I think its worth erroring out in a clean way as soon as possible.

hamilton/function_modifiers.py
elijahbenizzy: Yeah if we really want we can add future parameters starting with `_` (and block that now just to be 100% sure). IMO I think its fine given that we don't have any plans to add more params.

hamilton/function_modifiers.py
skrawcz: done.

elijahbenizzy (approved):
LGTM!

hamilton/function_modifiers.py
```diff
@@ -122,6 +129,93 @@ def validate(self, fn: Callable):
...
+        """A function is invalid if it does not have the requested parameter.
+
+        :param fn: Function to validate against this annotation.
+        :raises: InvalidDecoratorException If the function does not have the requested parameter
+        """
+        signature = inspect.signature(fn)
+        func_param_name_set = set(signature.parameters.keys())
+        try:
+            for output_name, mappings in self.parametrization.items():
+                fn.__doc__.format(**{self.RESERVED_KWARG: output_name}, **mappings)
```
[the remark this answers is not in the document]
skrawcz: added

PR #117, opened 2022-04-20
Adds some basic graph query functions to driver

skrawcz:
Thinking of how to use tags, or the workflow to use them, I'm envisioning

1. query for all nodes.
2. filter by tags.
3. use those nodes then for other purposes.

Ideally we could instead provide an API that you use tags to
query by instead, but I don't have a good feel for what that
should be right now.

Anyway, I'm thinking it'd be useful to ask questions,
given a set of function names, what is downstream or
upstream of them.

I only added visualization of the downstream because,
you can visualize what's upstream by using the
visualize_execution function.

## Changes

- Adds some driver functions for asking what is upstream and downstream of some functions.

## Testing

1. Worked locally.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -209,42 +209,42 @@ class Driver(object):
         nodes, user_nodes = self.graph.get_upstream_nodes(final_vars)
         return self.graph.has_cycles(nodes, user_nodes)
 
-    def what_is_downstream_of(self, *function_names: str) -> List[Variable]:
+    def what_is_downstream_of(self, *node_names: str) -> List[Variable]:
         """Tells you what is downstream of this function(s), i.e. node(s).
 
-        :param function_names: names of function(s) that are starting points for traversing the graph.
+        :param node_names: names of function(s) that are starting points for traversing the graph.
         :return: list of "variables" (i.e. nodes), inclusive of the function names, that are downstream of the passed
                 in function names.
         """
-        downstream_nodes = self.graph.get_impacted_nodes(list(function_names))
+        downstream_nodes = self.graph.get_impacted_nodes(list(node_names))
         return [Variable(node.name, node.type, node.tags) for node in downstream_nodes]
 
-    def display_downstream_of(self, *function_names: str, output_file_path: str, render_kwargs: dict):
+    def display_downstream_of(self, *node_names: str, output_file_path: str, render_kwargs: dict):
         """Creates a visualization of the DAG starting from the passed in function name(s).
 
         Note: for any "node" visualized, we will also add its parents to the visualization as well, so
         there could be more nodes visualized than strictly what is downstream of the passed in function name(s).
 
-        :param function_names: names of function(s) that are starting points for traversing the graph.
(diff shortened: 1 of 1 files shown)


hamilton/driver.py
```diff
@@ -209,6 +209,44 @@ def has_cycles(self, final_vars: List[str]) -> bool:
...
+        :param render_kwargs: a dictionary of values we'll pass to graphviz render function. Defaults to viewing.
+            If you do not want to view the file, pass in `{'view':False}`.
+        """
+        downstream_nodes = self.graph.get_impacted_nodes(list(function_names))
+        try:
+            self.graph.display(downstream_nodes, set(), output_file_path, render_kwargs=render_kwargs)
+        except ImportError as e:
+            logger.warning(f'Unable to import {e}', exc_info=True)
+
+    def what_is_upstream_of(self, *function_names: str) -> List[Variable]:
```
elijahbenizzy, line 240: Don't use function_name -- it won't be a 1:1 mapping with node name

elijahbenizzy (commented):
Honestly this feels a bit like the driver is getting bloated -- but I see the value of having these. At some point do we want to expose the underlying function_graph?

hamilton/driver.py
skrawcz: Agreed, but from the user perspective, they won't be thinking nodes, they will be thinking function names, right?

skrawcz:
> Honestly this feels a bit like the driver is getting bloated -- but I see the value of having these. At some point do we want to expose the underlying function_graph?

Yes/no. I think we want some API layer above the FunctionGraph class and not expose it directly. Whether this should be in the driver, sure, maybe not. To me, it should be in the driver if you're going to be using it close to where driver code lives. If you're just going to be building a driver to ask questions of the graph, and not execute anything, then perhaps a different object + interface would indeed be a better fit, as we could tailor that more rather than tacking it onto the driver as it is now.

hamilton/driver.py
elijahbenizzy: I mean I think this can be quite misleading. But yeah, maybe we want another name for it? Happy with node_name...

Also if we say `upstream`, then we're implying nodes in a DAG.

elijahbenizzy:
> > Honestly this feels a bit like the driver is getting bloated -- but I see the value of having these. At some point do we want to expose the underlying function_graph?
> 
> Yes/no. I think we want some API layer above the FunctionGraph class and not expose it directly. Whether this should be in the driver, sure, maybe not. To me, it should be in the driver if you're going to be using it close to where driver code lives. If you're just going to be building a driver to ask questions of the graph, and not execute anything, then perhaps a different object + interface would indeed be a better fit, as we could tailor that more rather than tacking it onto the driver as it is now.

Yeah I thin we want to see how its used more. Fine with this for now though.

elijahbenizzy (approved):
Happy as-is, but a few minor things that could change if you want.

PR #120, opened 2022-04-21
Ensures all functions have module tag defined and propagated

skrawcz:
We should add default tags to all functions when appropriate. I think
one of the appropriate things to do, is to include the module that the
function is in. This will be useful for people who create a DAG from
multiple modules and want some ability to find/query/subset functions.

I didn't add any other default ones, because I think we can think
about what they should be a little more. E.g. does it make sense for
extract_* to mention that the node created was done via the decorator? To
me it doesn't seem like we need that. When we add data quality, I think
there will be more of a use case there to tag "nodes" that are what data
quality added for instance.

## Changes

- Adds 'module' as a default tag.
- Makes any decorators/node creation propagate tags.

## Testing

1. Unit tests work.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7


hamilton/function_modifiers.py
```diff
@@ -380,7 +388,7 @@ def extractor_fn(field_to_extract: str = field, **kwargs) -> field_type:  # avoi
                 return kwargs[node_.name][field_to_extract]
 
             output_nodes.append(
-                node.Node(field, field_type, doc_string, extractor_fn, input_types={node_.name: dict}))
+                node.Node(field, field_type, doc_string, extractor_fn, input_types={node_.name: dict}, tags=node_.tags.copy()))
```
elijahbenizzy, line 391: Not sure this is the right call -- do we want to always add the tags from the node? For every field?

hamilton/function_modifiers.py
```diff
@@ -452,17 +460,19 @@ def validate(self, fn: Callable):
     def generate_node(self, fn: Callable, config) -> node.Node:
         """
         Returns one node which has the replaced functionality
-        :param config:
         :param fn:
+        :param config:
         :return:
         """
         fn_signature = inspect.signature(fn)
+        module_name = inspect.getmodule(fn).__name__
```
elijahbenizzy, line 468: Would be nice to separate this out into a utility function

hamilton/node.py
```diff
@@ -126,11 +126,14 @@ def depended_on_by(self) -> List['Node']:
...
 
+    def add_tag(self, tag_name: str, tag_value: str):
+        self._tags[tag_name] = tag_value
+
     def __hash__(self):
         return hash(self._name)
 
     def __repr__(self):
-        return f'<{self._name}>'
+        return f'<{self._name} {self._tags}>'
```
elijahbenizzy, line 136: Blech not sure if this is the best way to view these nodes... 

Also will this mess with some examples?

hamilton/function_modifiers.py
elijahbenizzy: E.G. `get_default_tags` and `get_module_from_function`

elijahbenizzy (commented):
Some minor changes

hamilton/function_modifiers.py
skrawcz:
```python
@tags(pii=True)
@extract_columns(...)
def load_df(...) -> pd.DataFrame:
```
If you read this, wouldn't you expect them to propagate?

Also this should result in the same outcome too:
```python
@extract_columns(...)
@tags(pii=True)
def load_df(...) -> pd.DataFrame:
```

hamilton/node.py
skrawcz: It's not a helpful `repr` in the first place -- so I think adding tags isn't bad.

No it wont affect any examples because this is on Node.

hamilton/function_modifiers.py
skrawcz: hmm, maybe. I'll sleep on it.

elijahbenizzy (approved):
LGTM -- make utility functions if you feel like it

hamilton/function_modifiers.py
skrawcz: done.

PR #125, opened 2022-05-01
Exposes passing kwargs to graphviz object

skrawcz:
This should be a backwards compatible API change for everything.

To be able to control, for instance the ratio, of the produced image,
it's not a render time argument. Instead, it's a property of the graph.
To be able to manipulate it, we need to wire through passing in attributes.
e.g. https://graphviz.org/doc/info/attrs.html shows what could be set.

These attributes are either graph, node, or edge attributes.
The kwargs I'm specifically trying to expose are:
```
graph_attr – Mapping of (attribute, value) pairs for the graph.
node_attr – Mapping of (attribute, value) pairs set for all nodes.
edge_attr – Mapping of (attribute, value) pairs set for all edges.
```
The above was taken from https://graphviz.readthedocs.io/en/stable/api.html?highlight=ratio#digraph.

Note, this will  enable anything else to be also passed to the graph, which I
think is a useful thing too. Since that will allow someone who knows graphviz
to really customize it I think.

## Changes

- Adds a `graphviz_kwargs` parameter and wires it through to all the displays.

## Testing

1. Ran local hello world script passing in `graphviz_kwargs`.

## Checklist

### Python - local testing

- [ ] python 3.6
- [X] python 3.7

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -165,16 +165,20 @@ class Driver(object):
         """
         return [Variable(node.name, node.type, node.tags) for node in self.graph.get_nodes()]
 
-    def display_all_functions(self, output_file_path: str, render_kwargs: dict = None):
+    def display_all_functions(self, output_file_path: str, render_kwargs: dict = None, graphviz_kwargs: dict = None):
         """Displays the graph of all functions loaded!
 
         :param output_file_path: the full URI of path + file name to save the dot file to.
             E.g. 'some/path/graph-all.dot'
         :param render_kwargs: a dictionary of values we'll pass to graphviz render function. Defaults to viewing.
             If you do not want to view the file, pass in `{'view':False}`.
+            See https://graphviz.readthedocs.io/en/stable/api.html#graphviz.Graph.render for other options.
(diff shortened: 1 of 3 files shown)


hamilton/driver.py
```diff
@@ -165,24 +165,28 @@ def list_available_variables(self) -> List[Variable]:
...
         except ImportError as e:
             logger.warning(f'Unable to import {e}', exc_info=True)
 
     def visualize_execution(self,
                             final_vars: List[str],
                             output_file_path: str,
                             render_kwargs: dict,
-                            inputs: Dict[str, Any] = None):
+                            inputs: Dict[str, Any] = None,
+                            graphviz_kwargs: dict = None):
```
elijahbenizzy, line 189: Can you explain the difference between this and `**render_kwargs`? I think it makes sense but its not obvious.

elijahbenizzy (approved):
Just make sure to clarify when one might use this as opposed to render_kwargs

hamilton/driver.py
skrawcz: Render is called after you construct the graph and want to create an image from it.

https://graphviz.readthedocs.io/en/stable/api.html#graphviz.Graph.render

PR #127, opened 2022-05-01
Remove unnecessary if statement

jameslamb:
Proposes removing an `if` statement in `FunctionGraph.display()` that is guaranteed to always evaluate to `True`.

## Changes

- Removes an `if` statement in `FunctionGraph.display()`

## Testing

N/A

## Notes

I'm confident the check being removed in this PR would always evaluate to `True`, as there is no way for user code to modify the local variable `kwargs`  before its truthiness is checked.

https://github.com/stitchfix/hamilton/blob/3eac094847dfd6beff70f5d660924a98a90e8277/hamilton/graph.py#L311-L312

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7


skrawcz:
yeah I think I just fixed that. Want to take a look at what's on main now?

jameslamb:
ah yep, looks like this was fixed in #125! nice!

PR #133, opened 2022-05-30
Fixes case where optional user inputs broke computation

skrawcz:
The execute function gets all upstream nodes of the required node to compute.
This will mean that there will likely be "user input" nodes to cycle through. When
we were computing the DFS value for them, we would assume they were required.

To illustrate, if you had a function that had optional input, `baz` e.g.

```python

def foo(bar: int, baz: float = 1.0) -> float:
```
This meant that if you did not pass in a value for `baz`, and `baz`
was a user input, Hamilton would complain that a required node
was not provided. Even though it was not required for computation.

So to fix that, in execute any user node is now marked with `optional`.
I believe this is fine to do, because if this is not the case, there will be a
node in the graph that will have `baz` as a REQUIRED dependency, and
thus things will break appropriately.

To help with that, I also fixed and added some unit tests.

One unit test is to ensure that we don't remove passing in `None` values
as part of the kwargs to the function. Since that's what we do now, and this
was another way to fix this bug, which I think would be the wrong solution.

## Changes

- any user node, when we come to compute, from a top level is optional.

## Testing

1. Adds unit tests.

## Checklist

### Python - local testing

- [x] python 3.6
- [x] python 3.7


skrawcz:
Ray issue is due to https://github.com/ray-project/ray/issues/25282

elijahbenizzy:
Ok, I think I see what you're getting at here. i think its OK, but not 100% sure. I guess my question is why would anyone be asking for the user-provided input as an output? Feels a little wonky, but I don't see a great reason why not...

tests/resources/optional_dependencies.py
```diff
@@ -22,6 +22,23 @@ def g(e: int, f: int = _F) -> int:
...
 
+def i(h: int, f: int = _F) -> int:
+    """we will pass None to e and so don't want the code to break for the unit test."""
+    if h is None:
+        h = 10
+    return h + f
+
+
+def none_result() -> int:
+    """Function to show that we don't filter out the result."""
```
elijahbenizzy, line 33: Why would we be filtering our the results in the first place?

elijahbenizzy (approved):
I think this looks good -- if I fully understand. Don't 100% get the original use-case this was breaking though...

tests/resources/optional_dependencies.py
skrawcz: we weren't -- added test to ensure that this behavior doesn't change without this test breaking.

skrawcz:
> I think this looks good -- if I fully understand. Don't 100% get the original use-case this was breaking though...

added example breaking case to description.

elijahbenizzy:
> I think this looks good -- if I fully understand. Don't 100% get the original use-case this was breaking though...

Ok I think I understand. Surprised we didn't catch this the first time, and it feels a little hacky, but I think its a fine solution.

> > I think this looks good -- if I fully understand. Don't 100% get the original use-case this was breaking though...
> 
> added example breaking case to description.

Alright, yep, makes sense. Thanks!

tests/test_graph.py
```diff
@@ -560,12 +560,83 @@ def test_optional_execute(config, inputs, overrides):
...
+    assert result == {'a': 10}
+    # something passed, something returned via inputs
+    result = fg.execute([fg.nodes['b'], fg.nodes['a']], inputs={'a': 10}, overrides={})
+    assert result == {'a': 10}
+    # something passed, something returned via overrides
+    result = fg.execute([fg.nodes['b'], fg.nodes['a']], inputs={}, overrides={'a': 10})
+    assert result == {'a': 10}
+
+
+@pytest.mark.parametrize('node_order', [
```
elijahbenizzy, line 590: Also...

```python
from itertools import permutations
ifh_permutations = list(permutations("ihj"))
...
```

skrawcz: yeah I was too lazy to look that up.

PR #140, opened 2022-06-23
Data quality examples

chrisaddy:

## Changes

-

## Testing

1.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7


hamilton/data_quality/default_validators.py
```diff
@@ -46,38 +50,58 @@ def arg(cls) -> str:
...
     def name(cls) -> str:
         return 'data_in_range_validator'
 
     def description(self) -> str:
-        return f'Validates that the datapoint falls within the range ({self.range[0]}, {self.range[1]})'
-
-    def validate(self, data: pd.Series) -> base.ValidationResult:
-        min_, max_ = self.range
-        between = data.between(min_, max_, inclusive=True)
+        if self.upper is None:
```
elijahbenizzy, line 76: Nice -- I like this extension!

hamilton/data_quality/default_validators.py
```diff
@@ -89,9 +113,44 @@ def validate(self, data: pd.Series) -> base.ValidationResult:
...
+            passes=passes,
+            message=message,
+            diagnostics=out_range
+        )
+
+
 
 class DataInRangeValidatorPrimitives(BaseDefaultValidator):
-    def __init__(self, range: Tuple[numbers.Real, numbers.Real], importance: str):
+    def __init__(self, range: Tuple[Comparable, Comparable], importance: str):
```
elijahbenizzy, line 153: Nice! This is much cleaner.

hamilton/data_quality/default_validators.py
```diff
@@ -46,38 +50,58 @@ def arg(cls) -> str:
...
+
 
     @classmethod
     def arg(cls) -> str:
         return 'range'
 
     @classmethod
     def applies_to(cls, datatype: Type[Type]) -> bool:
-        return issubclass(datatype, pd.Series)  # TODO -- handle dataframes?
+        return issubclass(datatype, Union[pd.Series, pd.DataFrame])
```
elijahbenizzy, line 69: Pretty sure this doesn't work in python?

```
>>> issubclass(int, typing.Union[int, float])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/elijahbenizzy/.pyenv/versions/3.9.10/lib/python3.9/typing.py", line 722, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
TypeError: Subscripted generics cannot be used with class and instance checks
```

hamilton/data_quality/default_validators.py
```diff
@@ -46,38 +50,58 @@ def arg(cls) -> str:
...
-    def validate(self, data: pd.Series) -> base.ValidationResult:
-        min_, max_ = self.range
-        between = data.between(min_, max_, inclusive=True)
+        if self.upper is None:
+            return f'Validates that the datapoint >= {self.lower}'
+        elif self.lower is None:
+            return f'Validates that the datapoint <= {self.upper}'
+        return f'Validates that the datapoint falls within the range ({self.lower}, {self.upper})'
+
+    def validate(self, data: Union[pd.DataFrame, pd.Series]) -> base.ValidationResult:
```
elijahbenizzy, line 82: Dataframes are an interesting addition -- we may want to find a way of excluding/including columns...

elijahbenizzy (commented):
Looks great! Really appreciate the examples/additions. Small point on the `issubclass` call (although if it works for you please show me your incantations...).

There might be a unit test we could add too (E.G. for dataframes?). 

Also we'll want to think through dataframes as we'll want a way of filtering for specific columns...

hamilton/data_quality/default_validators.py
chrisaddy: yea, I purposely haven't included any extension to dataframes for the datatype arg, in the example I am assuming you are dealing with a dataframe of a single dtype, which I think would probably be the most common use case? in this example, I just want to check on the correlation matrix, I think for any sort of "main path" in the DAG, you'd be splitting up the columns anyway -- the more hamiltonian way

though I think having a way to subset the columns could be useful, then the logic would be all the same more or less just would start the validator with 

```python
data = data[self.columns]
```

hamilton/data_quality/default_validators.py
chrisaddy: ah, yea, good point, need to add some tests to this draft anyway, so will make sure that gets cleaned up

hamilton/data_quality/default_validators.py
```diff
@@ -134,17 +182,218 @@ def validate(self, data: numbers.Real) -> ValidationResult:
...
+        importance: str,
+        available_validators: List[Type[BaseDefaultValidator]] = None,
+        **default_validator_kwargs) -> List[BaseDefaultValidator]:
+    """Resolves default validators given a set pof parameters and the type to which they apply.
+    Note that each (kwarg, type) combination should map to a validator
+    @param importance: importance level of the validator to instantiate
+    @param output_type: The type to which the validator should apply
+    @param available_validators: The available validators to choose from
+    @param default_validator_kwargs: Kwargs to use
+    @return: A list of validators to use
```
skrawcz, line 385: Formatting nit: we're using the `:param ...:` style rather than `@param ...`.

skrawcz:
Good start @chrisaddy!

I like seeing check_output applied to a real example. What it reminded me of, is our discussion we had about ensuring configurability of checks. Since specifying `importance` might be overkill on all functions/you will want a way to override them without updating the code - we should decide whether that's a blocker or follow up feature CC @elijahbenizzy .

Otherwise would love the PR to be broken out into more commits, where commits state the "why" behind some of the decisions you're making -- will make it easier to follow some of what you're trying to do, and then in the future if we have to change something, we'd have the commit reasoning to help guide our thinking.

PR #145, opened 2022-06-30
Adds the ability to pass functions not defined in modules - fixes #134

### LINKED ISSUE #134: Add ability to provide functions directly to driver
**Is your feature request related to a problem? Please describe.**
Hard to develop with Hamilton on Google Collab.

From Herve M. via discord:
> working with Google Colab, as this requires to have you feature .py scripts imported on the Colab machine, or git cloning them, but with the risk to loose them when the machine is recollected by Google. Having your features in the main notebook would make them safe on Colab.

**Describe the solution you'd like**
We should enable a way to pass functions into the driver directly.

# Option 1 - pass directly

```python
from my_timeseries_package import (
    rolling_mean,
    rolling_stdev
)
import other_func_module

d = Driver({},  rolling_mean, rolling_stdev, other_func_module)
```
Assuming we go with our current behavior, you would not be able to pass in two functions with the same name like happens now. i.e. no using this to "override" a function definition.

## Pros
This is pretty simple to pick up and implement for a user.

## Cons
1. This requires changing code in the driver/function graph. 
2. The long term software impacts leave the possiblity of creating messy driver code, because functions are not curated.

# Option 2 - curate functions into "module" like objects.

E.g. 
```python
from my_timeseries_package import (
    rolling_mean,
    rolling_stdev
)
import other_func_module

temp_module = TempModule(rolling_mean, rolling_stdev)
d = Driver({}, temp_module, other_func_module)
```
Assuming we go with our current behavior, you would not be able to pass in two functions with the same name like happens now. i.e. no using this to "override" a function definition.

## Pros
1. Does not require changes in the driver/function graph code.
2. Leaves open the possibility of doing option 1 at a later date.
3. Forces people to curate functions -- which should make refactoring the code easier in the future, versus allowing uncurated groups of functions.

## Cons
1. Introduces a new concept a user needs to know about.

**Additional context**
See https://discordapp.com/channels/842466558483496997/842467113611821087/982409991547940934 
Also from private chatter between @jameslamb and @elijahbenizzy.

skrawcz:
We went with the decision to create an object to house functions for three reasons:

1. It meant we did not need to structurally change the underlying code, other than checking
for the object instance type, and amending some type annotations.
2. This adds some friction to force people to at least acknowledge they need to curate the functions.
The name `TemporaryFunctionModule` should indicate that this should not be a production thing.
3. This opens the door for us to change our mind in the future about how we do this.

The `my_notebook_script.py` shows what one would do in a notebook context,
i.e. shows how you can inline define functions that interact with a module that constructs a DAG.

Assumption we did not break:

 - you cannot define the same function twice -- so you cannot use this as a way to override a  previously
defined function.

Adding this for context, if you want to overwrite a function implementation you instead have the following paths:

1. Choose the subset of functions you want and put them in the TemporaryFunctionModule, i.e. import those functions.
2. (a)Run the upstream part of the DAG to create the inputs required for the new function.
(b) Then compute the result with the new function. (c) Then run the rest of the DAG using the `override=` option of execute.

Usage example -- not showing imports:
```python
def signups() -> pd.Series:
    """Returns sign up values"""
    return pd.Series([1, 10, 50, 100, 200, 400])

def spend() -> pd.Series:
    """Returns the spend values"""
    return pd.Series([10, 10, 20, 40, 40, 50])

def log_spend_per_signup(spend_per_signup: pd.Series) -> pd.Series:
    """Simple function taking the logarithm of spend over signups."""
    return np.log(spend_per_signup)

temp_module = ad_hoc_utils.create_temporary_module(spend, signups, log_spend_per_signup, module_name='function_example')

initial_config = {}
dr = driver.Driver(initial_config, my_functions, temp_module)  # can pass in multiple modules
output_columns = [
    'spend',
    'signups',
    'avg_3wk_spend',
    'spend_per_signup',
    'spend_zero_mean_unit_variance',
    'log_spend_per_signup'
]
df = dr.execute(output_columns)
```

## Changes

- Adds module for ad_hoc_utils that enables one to create a temporary module.
- Adds tests & example showing how to make use of ad_hoc_utils.

## Testing

1. Ran it locally.

## Checklist

### Python - local testing

- [x] python 3.6
- [x] python 3.7

### DIFF
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -6,7 +6,6 @@ import abc
 import collections
 import inspect
 import typing
-from typing import Callable
 
 import pandas as pd
 import numpy as np
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -167,23 +167,3 @@ class SimplePythonGraphAdapter(SimplePythonDataFrameGraphAdapter):
     def build_result(self, **outputs: typing.Dict[str, typing.Any]) -> typing.Any:
         """Delegates to the result builder function supplied."""
         return self.result_builder.build_result(**outputs)
-
-
-class TemporaryFunctionModule(object):
-    """This class represents a way to encapsulate a set of functions for DAG construction.
-
-    It is meant to be used in a notebook, e.g. google collab, like context where you are iterating and prototyping.
-
-    Once you're happy with the functions you have created, you're supposed to transition the functions into a
-    proper python module.
-    """
-    __name__ = 'TemporaryFunctionModule'
-
-    def __init__(self, *functions: Callable):
-        """Constructor.
-
-        Takes in functions as arguments.
-        """
-        for function in functions:
-            # this sets each function as an attribute on this object.
-            setattr(self, function.__name__, function)
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -6,6 +6,7 @@ import abc
 import collections
 import inspect
 import typing
+from typing import Callable
 
 import pandas as pd
 import numpy as np
@@ -166,3 +167,23 @@ class SimplePythonGraphAdapter(SimplePythonDataFrameGraphAdapter):
     def build_result(self, **outputs: typing.Dict[str, typing.Any]) -> typing.Any:
         """Delegates to the result builder function supplied."""
         return self.result_builder.build_result(**outputs)
+
+
+class TemporaryFunctionModule(object):
+    """This class represents a way to encapsulate a set of functions for DAG construction.
+
+    It is meant to be used in a notebook, e.g. google collab, like context where you are iterating and prototyping.
+
+    Once you're happy with the functions you have created, you're supposed to transition the functions into a
+    proper python module.
+    """
+    __name__ = 'TemporaryFunctionModule'
+
+    def __init__(self, *functions: Callable):
+        """Constructor.
+
+        Takes in functions as arguments.
+        """
+        for function in functions:
+            # this sets each function as an attribute on this object.
+            setattr(self, function.__name__, function)
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -9,7 +9,7 @@ import inspect
 import logging
 import typing
 from types import ModuleType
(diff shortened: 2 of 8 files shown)


hamilton/base.py
```diff
@@ -166,3 +167,23 @@ def __init__(self, result_builder: ResultMixin):
...
+    __name__ = 'TemporaryFunctionModule'
+
+    def __init__(self, *functions: Callable):
+        """Constructor.
+
+        Takes in functions as arguments.
+        """
+        for function in functions:
+            # this sets each function as an attribute on this object.
+            setattr(self, function.__name__, function)
```
elijahbenizzy, line 189: This is too clever and I don't like it -- instead we should just store a list of functions, and below we return them if its a tempModule.

hamilton/graph.py
```diff
@@ -94,17 +94,18 @@ def types_match(adapter: base.HamiltonGraphAdapter,
...
 
     This iterates through the function module and grabs all function definitions.
     :return: list of tuples of (func_name, function).
     """
 
     def valid_fn(fn):
-        return (inspect.isfunction(fn)
-                and not fn.__name__.startswith('_')
-                and is_submodule(inspect.getmodule(fn), function_module))
+        if inspect.isfunction(fn) and not fn.__name__.startswith('_'):
```
elijahbenizzy, line 105: No need to loop through the functions here for tempmodule. Just store them and `if isinstance(module, base.TemporaryFunctionModule)` then `return `module.functions`. Whole different code-path, no need for inspect.

elijahbenizzy: What this does is bad in case we want to add any metadata to `TemporaryFunctionModule` that doesn't start with underscores -- it'l just pick it up.

elijahbenizzy (commented):
Too clever!

hamilton/base.py
skrawcz: sure -- guess there is a more pythonic way to do this...

elijahbenizzy (approved):
LGTM, let's do a minor version bump.

PR #147, opened 2022-07-03
Data quality integration with pandera

elijahbenizzy:
Integrates pandera with data quality

## Changes

- Adds schema= option to default data validators

## Testing

See unit tests, integration tests.

## Notes

Added a new suite/location of tests for integrations. Happy to revisit this.

## Checklist

### Python - local testing

- [x] python 3.6
- [x] python 3.7


elijahbenizzy:
Fairly basic pandera integration, but I actually like it quite a bit. Remaining before merge:

- [ ] Documentation
- [ ] Polishing of the conditional imports stuff
- [ ] Another pass at improving diagnostics/serializing pandera checks (might as well use the yaml stuff they have, although its clunky)

hamilton/data_quality/default_validators.py
```diff
@@ -312,6 +314,19 @@ def name(cls) -> str:
 ]
 
 
+def _append_pandera_to_default_validators():
+    """Utility method to append pandera validators as needed"""
+    try:
+        import pandera_validators
+    except ModuleNotFoundError:
+        logger.info(f'Cannot import pandera from pandera_validators. Run pip install hamilton[pandera] if needed.')
+        return
```
skrawcz, line 323: This might be clearer?
```suggestion
    try:
        import pandera
    except ModuleNotFoundError:
        logger.info(f'Skipping import of pandera for pandera_validators. Run pip install hamilton[pandera] if needed.')
        return
    else:
        import pandera_validators
```

skrawcz:
Question: Could the default validators we have be implemented instead with Pandera, i.e. delegate to it? If so, then perhaps a capability to swap the default implementation would be useful?

elijahbenizzy:
> Question: Could the default validators we have be implemented instead with Pandera, i.e. delegate to it? If so, then perhaps a capability to swap the default implementation would be useful?
> 
> 

Yes they could -- made a note in the commute though. I guess we have to decide whether it's worth having multiple implementations? Or just reimplementing the default to reduce surface area.

hamilton/data_quality/default_validators.py
elijahbenizzy: Agreed, that is a little easier to read.

hamilton/data_quality/default_validators.py
```diff
@@ -312,6 +314,20 @@ def name(cls) -> str:
...
 
+def _append_pandera_to_default_validators():
+    """Utility method to append pandera validators as needed"""
+    try:
+        import pandera
+    except ModuleNotFoundError:
+        logger.info(f"Cannot import pandera from pandera_validators. Run pip install hamilton[pandera] if needed.")
+        return
+   from . import pandera_validators
+    AVAILABLE_DEFAULT_VALIDATORS.extend(pandera_validators.PANDERA_VALIDATORS)
```
skrawcz, line 289: is this white space off?

data_quality.md
```diff
@@ -36,6 +36,46 @@ The available default validators are listed in the variable `AVAILABLE_DEFAULT_V
...
+
+We've fully integrated data quality with [pandera](https://pandera.readthedocs.io/en/stable/)!
+
+Note thtat you have to have hamilton installed with the `pandera` extension. E.G.
+
+```bash
+pip install hamilton[pandera]
+```
+
+This is near-trivial to use. All you have to do is provide a pandera schema
```
skrawcz, line 49: We shouldn't say it's easy incase we make someone feel stupid. Suggestion : "This should not take you long to start using".

skrawcz: Also link to what a pandera schema is -- and say something like (don't know what a pandera schema is, please read up on it here..)

data_quality.md
```diff
@@ -36,6 +36,46 @@ The available default validators are listed in the variable `AVAILABLE_DEFAULT_V
...
+            ]),
+        },
+        index=pa.Index(int),
+        strict=True,
+    ))
+def dataframe_with_schema(...) -> pd.DataFrame:
+    ...
+```
+
+You can also do schema checks on series, using the `pa.SeriesSchema` feature!
```
skrawcz, line 77: should provide an example here too. people want to copy & paste.

skrawcz (changes requested):
I just want a couple of items to be added to the docs.

Also a circular import is a code smell. So I think putting that base class in base makes sense...

Otherwise man those class imports from modules look pretty 🤮 ... will clean them up when we reformat using black...

hamilton/data_quality/default_validators.py
elijahbenizzy: Weird its not showing up on mine

data_quality.md
skrawcz: Frustrated < Stupid IMO.

skrawcz (approved):

hamilton/data_quality/default_validators.py
skrawcz: You fixed it in a later commit.

PR #151, opened 2022-07-08
Adds Data Quality examples + fixes

skrawcz:
This PR does quite a few things -- it's goal is to ensure usability of data quality when it is released, and that there is a good example for people to reference.

Review by commits.

## Changes

- Changes names of validators to be consistent
- Changes how failures are computed -- now all are computed versus stopping at the first one.
- Fixes Pandera integration and ensures it operates over dask data types.
- Adds two examples to examples folder.

## Testing

1. Locally tested on python 3.8 with normal, dask, ray, and pandas on spark (3.3).

## Notes

* I have added a folder to the examples directory with two sub folders: one using Hamilton validators, the other using Pandera.
* Dask data types do not play nice with the current validators. We could do something akin to what Pandera does, but my vote is to just use Pandera if you're in that situation.

## Checklist

- [x] Changes are limited to a single goal (no scope creep) --> Yes, by commit.

### Python - local testing

- [x] python 3.8
~- [ ] python 3.7~
~- [ ] python 3.6~


examples/data_quality/simple/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## How to run
+Running the code involves installing the right python dependencies, and then executing the code.
+It is best practice to create a python virtual environment for each project/example; we omit showing that step here.
+
+The DAG created for each way of executing is logically the same, though it might involve different functions
+being executed in the case of `spark` and `dask`; the `@config.when*` decorators are used for this purpose.
+
+Note: `importance` is not specified in the `@check_output`decorators found in feature_logic.py. The default
+behavior is therefore invoked, which is to log a "warning" and not stop execution. If stopping execution is desired,
```
elijahbenizzy, line 46: This is relevant to other pieces, but I think default should be `fail`, and `warning` should be additional. People are going to ignore/not notice.

examples/data_quality/simple/data_loaders.py
```diff
@@ -0,0 +1,89 @@
...
+    df = pd.read_csv(location, sep=';')
+    # rename columns to be valid hamilton names -- and lower case it
+    df.columns = _sanitize_columns(df.columns)
+    # create proper index -- ID-Month-Day;
+    index = df['id'].astype(np.str) + '-' + df['month_of_absence'].astype(np.str) + '-' + df['day_of_the_week'].astype(np.str)
+    df.index = index
+    return df
+
+
+@config.when(execution='dask')
```
elijahbenizzy, line 48: OK so when did we start using `config.when` on execution environments? That feels wrong to me. It could easily push a pattern where people start building execution methods into the DAG which subsequently gives us less freedom to execute however we want.

I think a cleaner approach (and maybe for the next step) is to have a `dataloader` decorator that does exactly this. Imagine:

```python
@dataloader(engine='dask')
def raw_data__dask(location: str, block_size: str = '10KB') -> dask.DataFrame
    ...
```

Perhaps it functions exactly like this under the hood, but its a lot cleaner and allows us not to lie on the type outputs (pandas versus dask, which is super confusing).

examples/data_quality/simple/data_loaders.py
```diff
@@ -0,0 +1,89 @@
...
+    from dask import dataframe
+    df = dataframe.read_csv(location, sep=';', blocksize=block_size)
+    # rename columns to be valid hamilton names -- and lower case it
+    df.columns = _sanitize_columns(df.columns)
+    # create proper index -- ID-Month-Day;
+    df.index = df['id'].astype(str) + '-' + df['month_of_absence'].astype(str) + '-' + df['day_of_the_week'].astype(str)
+    return df
+
+
+@config.when(execution='spark')
```
elijahbenizzy, line 65: Ditto, but this one is really coalas...

examples/data_quality/simple/feature_logic.py
```diff
@@ -0,0 +1,179 @@
...
+    1 - first season
+    2 - second season
+    3 - third season
+    4 - fourth season
+    """
+    return pd.get_dummies(seasons, prefix='seasons')
+
+
+@config.when_in(execution=['spark'])
+def seasons_encoded__spark(seasons: pd.Series) -> pd.DataFrame:
```
elijahbenizzy, line 59: Blurgh this `import pyspark.pandas` is kind of annoying... 

What if the function just took in the pandas engine as a parameter?

hamilton/base.py
```diff
@@ -42,6 +42,13 @@ class PandasDataFrameResult(ResultMixin):
     def build_result(**outputs: typing.Dict[str, typing.Any]) -> pd.DataFrame:
         # TODO check inputs are pd.Series, arrays, or scalars -- else error
         # TODO do a basic index check across pd.Series and flag where mismatches occur?
+        if len(outputs) == 1:
+            value = list(outputs.values())[0]
```
elijahbenizzy, line 46: `value, = output.values()`

elijahbenizzy (commented):
Alright, overall cleaner but tons of feedback. IMO there are one or two short-term trade-offs that make things more confusing and don't get us all that much.

hamilton/data_quality/default_validators.py
```diff
@@ -56,6 +56,58 @@ def validate(self, data: pd.Series) -> base.ValidationResult:
...
+    def validate(self, data: pd.Series) -> base.ValidationResult:
+        na_values = data.isna()
+        valid_values = data.isin(self.values)
+        final_valid_values = valid_values | na_values  # data is in valid values, or is NA
+        counts = final_valid_values.value_counts().to_dict()
+        correct_values_count = counts.get(True, 0)
+        incorrect_values_count = counts.get(False, 0)
+        incorrect_values = data[~valid_values].value_counts().to_dict()
+        passes = incorrect_values_count == 0
+        message = (f'Series contains {correct_values_count} correct values,'
```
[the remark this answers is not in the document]
skrawcz: > _if it's short_

Yep -- will just print values for this version. We can always adjust it.

hamilton/data_quality/default_validators.py
```diff
@@ -95,6 +147,46 @@ def name(cls) -> str:
...
+        super(DataInValuesValidatorPrimitives, self).__init__(importance=importance)
+        self.values = frozenset(values_in)
+
+    @classmethod
+    def arg(cls) -> str:
+        return 'values_in'
+
+    @classmethod
+    def applies_to(cls, datatype: Type[Type]) -> bool:
+        return issubclass(datatype, numbers.Real) or datatype == str  # TODO support list, dict and typing.* variants
```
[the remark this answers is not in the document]
skrawcz: this is using the type annotations on a function right? so in which case it would be something like `np.int64`? 

```python
> issubclass(np.int64, numbers.Real) 
True
```

hamilton/data_quality/pandera_validators.py
```diff
@@ -22,7 +22,9 @@ def description(self) -> str:
 
     def validate(self, data: pd.DataFrame) -> base.ValidationResult:
         try:
-            self.schema.validate(data, lazy=True)
+            result = self.schema.validate(data, lazy=True, inplace=True)
+            if hasattr(result, 'dask'):
```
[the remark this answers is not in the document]
skrawcz: We don't lazily evaluate it else where.

When this runs on vanilla code - that line evaluates things. It just so happens in the case of dask, that pandera just returns the dask object type thing. So we need to hit `.compute()` here, else we won't get what we want.

So I don't think this is leaking anything -- it's related to how Pandera handles it.

hamilton/data_quality/base.py
```diff
@@ -96,6 +96,7 @@ def act_fail_bulk(node_name: str, failures: List[Tuple[ValidationResult, DataVal
     for validation_result, validator in failures:
         if not validation_result.passes:
             message = f'{_create_error_string(node_name, validation_result, validator)}\n'
+            logger.error(message)  # log here so things print nicely at least
```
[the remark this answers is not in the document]
skrawcz: yes - show up nicely once.
Show up a second time as a very long string that might not even be printed nicely.

hamilton/base.py
skrawcz: done - squashed it into commit.

elijahbenizzy (approved):
LGTM let's release this

Issue #168, opened 2022-08-04
Scaling to 100K+ Nodes 


skrawcz:
# Current behavior
Creating a DAG with 100K+ nodes is slow. It took 1117.1986498832703 seconds for it to complete for the first example below.
Execution time was then 450 seconds for 100K colums at 10 rows deep running a validation check on them.

I then tried to do the same thing with Ray -- and it took longer, which makes sense since we're delegating over to Ray 100K+ tasks.
Interestingly Ray did not speed things up -- guess there's too much overhead here somewhere for the size of the task...

We should probably have a better way of handling creation of super large DAGs.

Use case: biology -- I came across this example in the Pandera discord channel where a user was trying to use pandera to validate a dataframe. Niels kindly provided some code to reproduce the issue, so I built on that here to see how Hamilton would fare. 

## Code to Reproduce
```python
import time
import random
import pandas as pd
import pandera as pa

from hamilton import driver, ad_hoc_utils, base
from hamilton.experimental import h_ray
from hamilton.function_modifiers import extract_columns, check_output

NUM_COLS = 100_000
cols = [f'i_{col}' for col in range(NUM_COLS)]

series_schema = pa.SeriesSchema(int, pa.Check.isin([0, 1, 2, 3]))


@check_output(schema=series_schema)
@extract_columns(*cols)
def data(number_of_columns: int) -> pd.DataFrame:
    df = pd.DataFrame.from_records([
        {
            f'i_{col}': random.choice([0, 1, 2, 3])
            for col in range(number_of_columns)
        }
        for _ in range(10)
    ])
    return df

if __name__ == '__main__':
    temp_module = ad_hoc_utils.create_temporary_module(data)
    start = time.time()
    print('Creating DAG')
    dr = driver.Driver({'number_of_columns': NUM_COLS}, temp_module)
    print(f"Created DAG in: {(time.time() - start)} seconds")
    df = dr.execute(cols)
    # dr.visualize_execution(cols, './viz.dot', {})
    # schema(data, lazy=True)
    print(f"Total Hamilton time: {(time.time() - start)} seconds")
    # print(df.head())

    del df
    del dr

    import ray
    ray.init()
    start = time.time()
    print('Creating RAY DAG')
    rga = h_ray.RayGraphAdapter(result_builder=base.SimplePythonDataFrameGraphAdapter())
    dr = driver.Driver({'number_of_columns': NUM_COLS}, temp_module, adapter=rga)
[...]

PR #169, opened 2022-08-04
Fixes DAG construction slowness

elijahbenizzy:
This is an issue with check_output. We were using sum(lists) which does
a concat operation for each list. This is fine at a small scale, but at
a larger scale this is an O(n^2) operation. We can still optimize this
significantly but this fixes the underlying issue in #168.

Note that this does *not* fix the issues around ray yet, this is just
the underlying piece. With Ray, the hard part is validating everything
which is inherently slow. Worth looking into why parallelism didn't help
though.

## Changes

-

## Testing

1.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7

### DIFF
--- a/hamilton/function_modifiers_base.py
+++ b/hamilton/function_modifiers_base.py
@@ -190,9 +190,11 @@ class NodeTransformer(SubDAGModifier):
 
         def node_tagged_non_final(node_: node.Node):
             return node_.tags.get(NodeTransformer.NON_FINAL_TAG, False)  # Defaults to final
-
-        non_final_nodes = set(
-            sum([[dep for dep in node_.input_types] for node_ in nodes], [node_.name for node_ in nodes if node_tagged_non_final(node_)]))
+        non_final_nodes = set()
+        for node_ in nodes:
+            for dep in node_.input_types:
+                if not node_tagged_non_final(node_):
+                    non_final_nodes.add(dep)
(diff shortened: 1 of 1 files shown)


skrawcz (approved):
I assume we have test coverage here? If so, LGTM.

skrawcz:
how much does it improve #168 ?

elijahbenizzy:
Need to run some actual benchmark but 7x for 10k nodes and 100x for 100k nodes.

elijahbenizzy:
Scales linearly:

```
Created DAG with 10 columns in: 0.0002918243408203125 seconds
Created DAG with 100 columns in: 0.0019080638885498047 seconds
Created DAG with 1000 columns in: 0.031607866287231445 seconds
Created DAG with 10000 columns in: 0.3109908103942871 seconds
Created DAG with 100000 columns in: 4.007328033447266 seconds
```

Compared to:
```
Created DAG with 10 columns in: 0.00030612945556640625 seconds
Created DAG with 100 columns in: 0.0021920204162597656 seconds
Created DAG with 1000 columns in: 0.05421900749206543 seconds
Created DAG with 10000 columns in: 2.8478641510009766 seconds
# don't feel like waiting
```

elijahbenizzy:
OK, so where does it spend its time?

image

No obvious bang for our buck -- @skrawcz your intuition is right once we fixed it getting *super* slow. Object creation takes 20=25% of the time. Then resolving arguments (I think we can probably cache that) takes another 230%. So, not big solutions here. Going to call this a success and merge.

PR #171, opened 2022-08-06
Async implementation of driver/adapter

elijahbenizzy:
@skrawcz this is what I meant

Basically we pass the coroutine the entire way through.
When a function depends on other data, we create a coroutine that
awaits everything. Finally, we do a gather at the end.

I *think* this is optimal but will need to dig in a bit/look into
things.

Otherwise, nifty POC with *very* rough edges.

The biggest hack is that I'm using a cache to store coroutine values as python makes it very hard to access them. 
I think we can do better, but the point is we don't have to mess with execution.

## Changes

-

## Testing

1.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7

### DIFF
--- a/examples/async/README.md
+++ b/examples/async/README.md
@@ -20,7 +20,7 @@ Then curl with:
 curl -X 'POST' \
   'http://localhost:8000/execute' \
   -H 'accept: application/json' \
-  -d ''
+  -d '{}'
 ```
 
 You should get the following result:
@@ -45,7 +45,7 @@ in external tasks as inputs if you want.
 1. This will break in certain cases when decorating an async function (E.G. with `extract_outputs`).
 This is because the output of that function is never awaited during delegation. We are looking into ways to fix this,
 but for now be careful. We will at least be adding validation so the errors are clearer.
-2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions
+2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions.
 
 We want feedback! We can determine how to make this part of the core API once we get userse who are happy,
 so have some fun!
--- a/examples/async/README.md
+++ b/examples/async/README.md
@@ -1,0 +1,58 @@
+# Hamilton + Async
+
+This is currently an experimental feature, allowing one to run a hamilton DAG composed (entirely or partially) of async functions.
+
+## How to use
+
+See the [example](fastapi.py) for the example. The difference from a normal driver is two-fold:
+
+1. You call it using the `AsyncDriver` rather than the standard driver
+2. `raw_execute`, and `execute` are both coroutines, meaning they should be called with `await`.
+3. It allows for coroutines as inputs -- they'll get properly awaited
+
+To run the example, make sure to install `requirements.txt`.
+
+Then run `uvicorn fastapi_example:app` in one terminal.
+
+Then curl with:
+
+```bash
+curl -X 'POST' \
+  'http://localhost:8000/execute' \
+  -H 'accept: application/json' \
+  -d ''
+```
+
+You should get the following result:
+
+```json
+{"pipeline":{"computation1":false,"computation2":true}}
+```
+
+
+## How it works
+
+Behind the scenes, we create a [GraphAdapter](../../hamilton/experimental/h_async.py)
+that turns every function into a coroutine. The function graph then executes, solely creating
+more coroutines, that are evaluated at the end. Thus no computation is done until a final node
+is awaited.
+
+Any node inputs are awaited on prior to node computation if they are coroutines.
+
+This actually caches the coroutine outputs so we calculate a node once, and stay far away from
+python's issues around asking an already-computed coroutine its result (which is near impossible).
+
+## Caveats
+
+1. This will break in certain cases when decorating an async function (E.G. with `extract_outputs`).
+This is because the output of that function is never awaited during delegation. We are looking into ways to fix this,
+but for now be careful. We will at least be adding validation so the errors are clearer.
+2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions
+3. Currently there is the potential for a memory leak. This is due to the coroutine cache we use,
+which we really shouldn't need to. That said, you'll want to create a driver on a per-request basis (for now)
+## Next steps
+
+We want feedback! We can determine how to make this part of the core API once we get userse who are happy,
+so have some fun!
+
+Fixing the caveats will be the first order of business, and adding validations when things won't work.
--- a/examples/async/README.md
+++ b/examples/async/README.md
@@ -8,7 +8,7 @@ See the [example](fastapi.py) for the example. The difference from a normal driv
 
 1. You call it using the `AsyncDriver` rather than the standard driver
 2. `raw_execute`, and `execute` are both coroutines, meaning they should be called with `await`.
-3. It allows for coroutines as inputs -- they'll get properly awaited
+3. It allows for tasks as inputs -- they'll get properly awaited
 
 To run the example, make sure to install `requirements.txt`.
 
@@ -33,14 +33,12 @@ You should get the following result:
 ## How it works
 
 Behind the scenes, we create a [GraphAdapter](../../hamilton/experimental/h_async.py)
-that turns every function into a coroutine. The function graph then executes, solely creating
-more coroutines, that are evaluated at the end. Thus no computation is done until a final node
+that turns every function into a coroutine. The function graph then executes, creating tasks for each node,
+that are awaited at the end. Thus no computation is complete until a final node
 is awaited.
 
-Any node inputs are awaited on prior to node computation if they are coroutines.
-
-This actually caches the coroutine outputs so we calculate a node once, and stay far away from
-python's issues around asking an already-computed coroutine its result (which is near impossible).
+Any node inputs are awaited on prior to node computation if they are awaitable, so you can pass
+in external tasks as inputs if you want.
 
 ## Caveats
 
@@ -48,9 +46,6 @@ python's issues around asking an already-computed coroutine its result (which is
 This is because the output of that function is never awaited during delegation. We are looking into ways to fix this,
 but for now be careful. We will at least be adding validation so the errors are clearer.
 2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions
-3. Currently there is the potential for a memory leak. This is due to the coroutine cache we use,
-which we really shouldn't need to. That said, you'll want to create a driver on a per-request basis (for now)
-## Next steps
 
 We want feedback! We can determine how to make this part of the core API once we get userse who are happy,
 so have some fun!
--- a/examples/async/fastapi_example.py
+++ b/examples/async/fastapi_example.py
@@ -1,0 +1,60 @@
+import asyncio
+
+import aiohttp
+
+import fastapi
+from hamilton import ad_hoc_utils
+from hamilton.experimental import h_async
+
+
+async def request_raw(request: fastapi.Request) -> dict:
+    return await request.json()
+
+
+def foo(request_raw: fastapi.Request) -> str:
+    return request_raw.get('foo', 'far')
+
+
+def bar(request_raw: fastapi.Request) -> str:
+    return request_raw.get('bar', 'baz')
+
+
+async def computation1(foo: str, some_data: dict) -> bool:
+    await asyncio.sleep(1)
+    return False
+
+
+async def some_data() -> dict:
+    async with aiohttp.ClientSession() as session:
+        async with session.get('http://httpbin.org/get') as resp:
+            return await resp.json()
+
+
+async def computation2(bar: str) -> bool:
+    await asyncio.sleep(1)
+    return True
+
+
+async def pipeline(computation1: bool, computation2: bool) -> dict:
+    await asyncio.sleep(1)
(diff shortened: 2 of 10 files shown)


hamilton/driver.py
```diff
@@ -267,6 +269,66 @@ def what_is_upstream_of(self, *node_names: str) -> List[Variable]:
...
+            return self.coroutine_cache[val_id]
+        output = await val
+        self.coroutine_cache[val_id] = output
+        return output
+
+    def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) -> typing.Any:
+        callabl = node.callable
+
+        async def new_fn(fn=callabl, **fn_kwargs):
+            fn_kwargs = {key: await self.process_value(value) for key, value in fn_kwargs.items()}
```
skrawcz, line 300: make this a gather and it'll then parallelize computation of dependencies?

hamilton/driver.py
```diff
@@ -267,6 +269,66 @@ def what_is_upstream_of(self, *node_names: str) -> List[Variable]:
...
+        self.coroutine_cache = {}  # Ughh python's coroutine API is terrible
+
+    @staticmethod
+    def check_input_type(node_type: typing.Type, input_value: typing.Any) -> bool:
+        # TODO -- check if its a coroutine
+        return True
+
+    @staticmethod
+    def check_node_type_equivalence(node_type: typing.Type, input_type: typing.Type) -> bool:
+        return True  # We should just delegate to the superclass -- that should have a default implementation...
```
skrawcz, line 284: can just delete these.

hamilton/driver.py
```diff
@@ -267,6 +269,66 @@ def what_is_upstream_of(self, *node_names: str) -> List[Variable]:
         return [Variable(node.name, node.type, node.tags) for node in upstream_nodes]
 
 
+class AsyncGraphAdapter(base.HamiltonGraphAdapter):
+
+    def __init__(self):
+        self.coroutine_cache = {}  # Ughh python's coroutine API is terrible
```
skrawcz, line 275: good practice to call super().__init__() in the constructor

hamilton/driver.py
```diff
@@ -267,6 +270,69 @@ def what_is_upstream_of(self, *node_names: str) -> List[Variable]:
         return [Variable(node.name, node.type, node.tags) for node in upstream_nodes]
 
 
+async def await_dict_of_tasks(task_dict: Dict[str, types.CoroutineType]) -> Dict[str, Any]:
+    """Util to await a dictionary of tasks as asyncio.gather is kind of garbage"""
+    keys = sorted(task_dict.keys())
+    coroutines = [task_dict[key] for key in keys]
+    coroutines_gathered = await asyncio.gather(*coroutines)
+    return {key: value for key, value in zip(keys, coroutines_gathered)}
```
skrawcz, line 278:
```suggestion
    return dict(zip(keys, coroutines_gathered))
```

hamilton/driver.py
elijahbenizzy: Nope as its abstract? That said, I changed it to inherit from the implemented one.

elijahbenizzy:
Remaining:
- [ ] Tests
- [ ] Determine how to release (experimental?)

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def build_result(self, **outputs: typing.Dict[str, typing.Any]) -> typing.Any:
+        """Currently this is a no-op -- it just delegates to the resultsbuilder.
+        That said, we *could* make it async, but it feels wrong -- this will just be
+        called after `raw_execute`.
+
+        @param outputs: Outputs (awaited) from the graph.
+        @return: The final results.
+        """
+        return self.result_builder.build_result(**outputs)
```
skrawcz, line 101: probably want to clear the cache here?

examples/async/fastapi_example.py
```diff
@@ -0,0 +1,60 @@
...
+        {},
+        ad_hoc_utils.create_temporary_module(
+            pipeline,
+            computation1,
+            foo,
+            bar,
+            some_data,
+            computation2))
+    input_data = {'request_raw': request}
+    return await dr.raw_execute(['pipeline'], inputs=input_data)
```
skrawcz, line 60: hmm --right now if this is constructed per request, we're good -- things should be cleaned up and there'd be no memory leaks. However that incurs a penalty for each request.

But if we wanted to construct the driver once, then I think we'd get a memory leak? Also if multiple requests are sharing the same driver object & adapter, would anything weird happen 🤔 ?

hamilton/experimental/h_async.py
elijahbenizzy: Yeah, that's a great point

examples/async/fastapi_example.py
elijahbenizzy: Yep, I think you're right here -- it could easily fill up. With multiple requests, it should be fine - they won't share the same set of coroutines. Even if they do (say you pass a coroutine in), the GIL assures parallelism works fine.

elijahbenizzy: The problem is if we clear, it could break it, so we should really determine a run ID and clear for that run (as you said above)

elijahbenizzy: I'd prefer to not have to manage this cache at all, as python should be able to do this, so I'm going to put this as a caveat for now, and see if I can get rid of this cache in the first place.

examples/async/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Any node inputs are awaited on prior to node computation if they are awaitable, so you can pass
+in external tasks as inputs if you want.
+
+## Caveats
+
+1. This will break in certain cases when decorating an async function (E.G. with `extract_outputs`).
+This is because the output of that function is never awaited during delegation. We are looking into ways to fix this,
+but for now be careful. We will at least be adding validation so the errors are clearer.
+2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions
```
skrawcz, line 48:
```suggestion
2. Performance *should* be close to optimal but we have not benchmarked. We welcome contributions.
```

examples/async/fastapi_example.py
```diff
@@ -0,0 +1,60 @@
...
+    """Handler for pipeline call"""
+    dr = h_async.AsyncDriver(
+        {},
+        ad_hoc_utils.create_temporary_module(
+            pipeline,
+            computation1,
+            foo,
+            bar,
+            some_data,
+            computation2))
```
skrawcz, line 58: We should probably not do this in an example. This doesn't feel very "ad hoc".

Otherwise 💡, fastapi could be a possible code compilation target too.

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+        This does *not* use async def, as we want it to be awaited on later -- this await is done
+        in processing parameters of downstream functions/final results. We can ensure that as
+        we also run the driver that this corresponds to.
+
+        Note that this assumes that everything is a coroutine, even if it isn't.
+        It just wraps it in one.
+
+        @param node: Node to wrap
+        @param kwargs: Keyword arguments (either coroutines or raw values) to call it with
+        @return: A coroutine
```
skrawcz, line 69: should this be task?

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @staticmethod
+    def check_input_type(node_type: Type, input_value: Any) -> bool:
+        """Checks the input type to ensure that it is compatible with the node type.
+        Note that this also allows coroutines, in case you want the DAG to await
+        the result of calling the input.
+
+        @param node_type: Type of the input node, expected
+        @param input_value: Type of the actual input, received
+        @return: Whether the actual input value matches the type expected
```
skrawcz, line 54: we don't use this style :)
```suggestion
        :param node_type: Type of the input node, expected
        :param input_value: Type of the actual input, received
        :return: Whether the actual input value matches the type expected
```

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+    coroutines_gathered = await asyncio.gather(*coroutines)
+    return dict(zip(keys, coroutines_gathered))
+
+
+async def process_value(val: Any) -> Any:
+    """Helper function to process the value of a potential awaitable.
+    This is very simple -- all it does is await the value if its not already resolved.
+
+    @param val: Value to process.
+    @return: The value (awaited if it is a coroutine, raw otherwise).
```
skrawcz, line 26:
```suggestion
    :param val: Value to process.
    :return: The value (awaited if it is a coroutine, raw otherwise).
```

hamilton/experimental/h_async.py
skrawcz, line 69:
```suggestion
        :param node: Node to wrap
        :param kwargs: Keyword arguments (either coroutines or raw values) to call it with
        :return: A task
```

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+        return self.result_builder.build_result(**outputs)
+
+
+class AsyncDriver(driver.Driver):
+    def __init__(self, config, *modules, result_builder: Optional[base.ResultMixin] = None):
+        """Instantiates an asynchronous driver.
+
+        @param config: Config to build the graph
+        @param modules: Modules to crawl for fns/graph nodes
+        @param result_builder: Results mixin to compile the graph's final results. TBD whether this should be included in the long run.
```
skrawcz, line 100:
```suggestion
        :param config: Config to build the graph
        :param modules: Modules to crawl for fns/graph nodes
        :param result_builder: Results mixin to compile the graph's final results. TBD whether this should be included in the long run.
```

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+                          inputs: Dict[str, Any] = None) -> Dict[str, Any]:
+        """Executes the graph, returning a dictionary of strings (node keys) to final results.
+
+        Note inputs can be coroutines as well if you desire.
+
+        @param final_vars: Variables to execute (+ upstream)
+        @param overrides: Overrides for nodes
+        @param display_graph: whether or not to display graph -- this is not supported.
+        @param inputs:  Inputs for DAG runtime calculation
+        @return: A dict of key -> result
```
skrawcz, line 117:
```suggestion
        :param final_vars: Variables to execute (+ upstream)
        :param overrides: Overrides for nodes
        :param display_graph: whether or not to display graph -- this is not supported.
        :param inputs:  Inputs for DAG runtime calculation
        :return: A dict of key -> result
```

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        nodes, user_nodes = self.graph.get_upstream_nodes(final_vars, inputs)
+        memoized_computation = dict()  # memoized storage
+        self.graph.execute(nodes, memoized_computation, overrides, inputs)
+        if display_graph:
+            raise ValueError(f'display_graph=True is not supported for the async graph adapter. '
+                             f'Instead you should be using visualize_execution.')
+        return await await_dict_of_tasks({key: asyncio.create_task(process_value(memoized_computation[key])) for key in final_vars})
```
skrawcz, line 125:
```suggestion
        task_dict = {key: asyncio.create_task(process_value(memoized_computation[key])) for key in final_vars}
        return await await_dict_of_tasks(task_dict)
```
Also is `process_value` needed here? Seems a bit redundant with the result being wrapped in a task and that awaited?

examples/async/fastapi_example.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+async def computation2(bar: str) -> bool:
+    await asyncio.sleep(1)
+    return True
+
+
+async def pipeline(computation1: bool, computation2: bool) -> dict:
+    await asyncio.sleep(1)
+    return {'computation1': computation1, 'computation2': computation2}
```
skrawcz, line 40: let's move these to a proper module.

skrawcz (changes requested):

hamilton/experimental/h_async.py
elijahbenizzy: Ughh need to update my pycharm to auto-correct to the right one

hamilton/experimental/h_async.py
elijahbenizzy: `process_value` is needed but `create_task` is not I think...

skrawcz: okay. But this seems like an important line that should be at least two... break out the dict comprehension at least.

hamilton/experimental/h_async.py
```diff
@@ -0,0 +1,151 @@
...
+        In that case, it just wraps it in one.
+
+        :param node: Node to wrap
+        :param kwargs: Keyword arguments (either coroutines or raw values) to call it with
+        :return: A task
+        """
+        callabl = node.callable
+
+        async def new_fn(fn=callabl, **fn_kwargs):
+            fn_kwargs = await await_dict_of_tasks({key: process_value(value) for key, value in fn_kwargs.items()})
```
skrawcz, line 74: can move this dict comprehension to it's own line.

skrawcz (approved):
Just those two code nits, otherwise I think this looks good.

PR #173, opened 2022-08-08
Adds union support check when passing in inputs

### LINKED ISSUE #170: ValueError with typing.Union type signature
`hamilton` raises a `ValueError` when using `typing.Union`, even if the type signature has the input type.

# Current behavior
Error when using `typing.Union` types.

## Stack Traces
```
ValueError: 1 errors encountered:
  Error: Type requirement mismatch. Expected x:typing.Union[int, pandas.core.series.Series] got 1 instead.
```

## Steps to replicate behavior
`run.py`
```python
from hamilton.driver import Driver
from hamilton.base import SimplePythonGraphAdapter, DictResult
import lib

initial_columns = {'x': 1}
adapter = SimplePythonGraphAdapter(DictResult)
dr = Driver(initial_columns, lib, adapter=adapter)
print(dr.execute(['foo_union']))
```

`lib.py`
```python
import typing as t
import pandas as pd


def foo_union(x: t.Union[int, pd.Series]) -> t.Union[int, pd.Series]:
    '''foo union'''
    return x + 1
```

## Library & System Information
python 3.6, hamilton 1.9.0, tested on mac and linux.

# Expected behavior
To be able to run dag.

# Additional context
Everything works if there are functions that have `typing.Any` as type signature. For example, the above will run if we add `foo_any` to `lib.py`:

```python
import typing as t
import pandas as pd


def foo_union(x: t.Union[int, pd.Series]) -> t.Union[int, pd.Series]:
    '''foo union'''
    return x + 1


def foo_any(x: t.Any) -> t.Any:
    '''foo any'''
    return x + 1
```

skrawcz:
This should fix #170. If an input function was annotated
with a union, we would barf. Now it wont.

This still means downstream nodes need to have the same
type signature -- this doesn't change any behavior there.

## Changes

- adds support to expand union type when checking inputs

## Testing

1. Unit tests work.

## Checklist

### Python - local testing

- [x] python 3.6
- [x] python 3.7


skrawcz:
This code works
```python
import typing as t
import pandas as pd
def foo_union(x: t.Union[int, pd.Series]) -> t.Union[int, pd.Series]:
    '''foo union'''
    return x + 1
from hamilton.driver import Driver
from hamilton.base import SimplePythonGraphAdapter, DictResult
from hamilton import ad_hoc_utils
lib = ad_hoc_utils.create_temporary_module(foo_union)
initial_columns = {'x': 1}
adapter = SimplePythonGraphAdapter(DictResult)
dr = Driver(initial_columns, lib, adapter=adapter)
print(dr.execute(['foo_union']))
```

hamilton/base.py
```diff
@@ -150,6 +150,9 @@ def check_input_type(node_type: typing.Type, input_value: typing.Any) -> bool:
             return True
         elif typing_inspect.is_generic_type(node_type) and typing_inspect.get_origin(node_type) == type(input_value):
             return True
+        elif typing_inspect.is_union_type(node_type):
+            union_types = typing_inspect.get_args(node_type)
+            return any([SimplePythonDataFrameGraphAdapter.check_input_type(ut, input_value) for ut in union_types])
```
elijahbenizzy, line 155: I think we can add this under `check_node_type_equivalence` as well. As it currently stands, `Union[str, int]` will not be a valid upstream input *or* downstream input of `str` or `int`, but I would argue that it *should* be a valid downstream input.

hamilton/base.py
```diff
@@ -159,7 +159,13 @@ def check_input_type(node_type: typing.Type, input_value: typing.Any) -> bool:
 
     @staticmethod
     def check_node_type_equivalence(node_type: typing.Type, input_type: typing.Type) -> bool:
-        return node_type == input_type
+        if node_type == input_type:
+            return True
+        elif typing_inspect.is_union_type(input_type):
+            union_types = typing_inspect.get_args(input_type)
+            return any([SimplePythonDataFrameGraphAdapter.check_node_type_equivalence(node_type, ut)
```
elijahbenizzy, line 166: FYI this is why this should be a `self` method. In case we have unions of datatypes that would work for dask, this will break. Now, I'm not sure we'll need that, but it should at least be consistent. Now's a good time to change it -- it should all be backwards compatible as long as there aren't too many static-style calls (like above).

skrawcz: Yep -- I think I made them static because they didn't reference self. Will think about it.

I don't understand this comment:

>  In case we have unions of datatypes that would work for dask, this will break.

elijahbenizzy: It doesn't delegate to the dask graph adapter's union. Meaning that, if we have a union of two types that would work for the dask graph adapter but would break for the simple adapter, this will return False, as it delegates to the simple adapter's type-compatibility method regardless of the calling class. While this is probably a strange edge-case, it would be solved by utilizing `self` and recursively calling `self`'s compatibility checker.

elijahbenizzy:
OK, I just realized, this is really *not* the right place to put this. Instead it should live here: https://github.com/stitchfix/hamilton/blob/15271fa33bd47e5eb62209da357cfcb7d2243f8b/hamilton/graph.py#L30. It doesn't have anything to do with a graph adapter... instead it has to do with forming the DAG. Let's move it over?

hamilton/base.py
skrawcz, line 155: this can be removed, now right?

hamilton/data_quality/default_validators.py
```diff
@@ -374,6 +374,37 @@ def arg(cls) -> str:
...
+    @classmethod
+    def applies_to(cls, datatype: Type[Type]) -> bool:
+        return True
+
+    def description(self) -> str:
+        if self.allow_none:
+            return 'No-op validator.'
+        return 'Validates that an output ;is not None'
+
+    def validate(self, data: Any) -> ValidationResult:
```
skrawcz, line 392:
```suggestion
    def validate(self, data: Any) -> base.ValidationResult:
```

hamilton/data_quality/default_validators.py
```diff
@@ -374,6 +374,37 @@ def arg(cls) -> str:
...
+        passes = True
+        if not self.allow_none:
+            if data is None:
+                passes = False
+        return ValidationResult(
```
skrawcz, line 397:
```suggestion
        return base.ValidationResult(
```

hamilton/data_quality/default_validators.py
```diff
@@ -6,7 +6,7 @@
 import pandas as pd
 
 from hamilton.data_quality import base
-from hamilton.data_quality.base import BaseDefaultValidator
+from hamilton.data_quality.base import BaseDefaultValidator, ValidationResult
```
skrawcz, line 9:
```suggestion
from hamilton.data_quality.base import BaseDefaultValidator
```

hamilton/base.py
elijahbenizzy: No, this should remain. We do this in the graphadapter -- see comment here: https://github.com/stitchfix/hamilton/blob/ca529c1d9010f512af794eed5e0a804eea609b2c/hamilton/driver.py#L144

hamilton/data_quality/default_validators.py
elijahbenizzy: Out of scope

hamilton/base.py
skrawcz: ah yes -- makes sense. This should remain.

skrawcz:
LGTM  - you need to approve @elijahbenizzy

elijahbenizzy (approved):

PR #199, opened 2022-09-19
Subdag modifier

elijahbenizzy:

## Changes

-

## Testing

1.

## Checklist

### Python - local testing

- [ ] python 3.6
- [ ] python 3.7


elijahbenizzy:
Purely a WIP right now, need to finish implementation then add a ton of tests

hamilton/function_modifiers_base.py
```diff
@@ -1,363 +1,4 @@
...
-    for dag_modifier in node_transformers:
-        nodes = dag_modifier.transform_dag(nodes, config, fn)
-    node_decorators = getattr(fn, NodeDecorator.get_lifecycle_name(), [DefaultNodeDecorator()])
-    for node_decorator in node_decorators:
-        nodes = node_decorator.transform_dag(nodes, config, fn)
-    return nodes
+# Quick hack to make this a backwards compatible refactor
+# This allows us to define everything within the function_modifiers directory, and just refer to that
+# While maintaining old imports
+from hamilton.function_modifiers.function_modifiers_base import *
```
skrawcz, line 4: who imports this module?

Otherwise does this change preserve commit history? Would prefer for example to move this file and then add this file -- maybe in separate commits.

tests/test_function_modifiers.py
```diff
@@ -5,20 +5,20 @@
 import pandas as pd
 import pytest
 
+import hamilton.function_modifiers.function_modifiers_base
 from hamilton import function_modifiers, function_modifiers_base, models, node
```
skrawcz, line 9: should replace these imports no? at least the `function_modifiers_base` ?

skrawcz (commented):
I'm concerned about:

1. commit history preservation - wouldn't want to lose some of the "why" for some of the decorators.
2. Any reason function_modifiers_base needs to exist? Where is it being used? Can we add a deprecation warning if we can't migrate things ourselves?
3. Shouldn't there be a `base.py` in the function_modifiers module now?

elijahbenizzy:
> I'm concerned about:
> 
> 1. commit history preservation - wouldn't want to lose some of the "why" for some of the decorators.
> 2. Any reason function_modifiers_base needs to exist? Where is it being used? Can we add a deprecation warning if we can't migrate things ourselves?
> 3. Shouldn't there be a `base.py` in the function_modifiers module now?


1. Yeah, that's tough, but IMO the fact that they're all stuck together is worse. Did `git mv` but git can't support a refactor like this. Commit history will never be lost, just harder to find... Can also compile a document with all the changelog and attach it.
2. I think this was to support a Stitch Fix use-case, but it should not need to exist. Can definitely add a deprecation warning. Its not supposed to be a public API.
3. Yeah we can change that from `function_modifiers_base.py` to `base.py`

elijahbenizzy:
Some qs this brings up:
1. Should we have the `MultiOutput` as the decorating node?  Its currently duplicated for readability, but it feels unecessary.
2. How should we handle namespaces? I've made it part of the node name, but we could also give the node a `namespace` field, and use that to form the node name.
3. How should we handle multiple parameterizations of this? Might be worth thinking through, as that's going to be the first question people have when they see this. And it should be easy -- one could provide a mapping of namespaces to reuse_subdag objects, or something similar... All on the same function.

skrawcz:
@elijahbenizzy we can close this PR now right?

elijahbenizzy:
This is very odl and has been merged/implemented in its final form: https://github.com/stitchfix/hamilton/pull/317

PR #211, opened 2022-10-17
Makes the fastapi async example more robust

skrawcz:
When you use `/docs` to create the request, it does not pass in an empty dict, so making sure the path does not break.

## Changes

- adds try/except to async example
- adds way to run fastapi by just running `fastapi_example.py`

## Testing

1. Works locally

## Checklist

### Python - local testing

- [x] python 3.9

### DIFF
--- a/examples/async/async_module.py
+++ b/examples/async/async_module.py
@@ -1,11 +1,19 @@
 import asyncio
+import json
+import logging
 
 import aiohttp
 import fastapi
 
+logger = logging.getLogger(__name__)
+
 
 async def request_raw(request: fastapi.Request) -> dict:
-    return await request.json()
+    try:
+        return await request.json()
+    except json.JSONDecodeError as e:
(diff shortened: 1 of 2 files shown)


examples/async/async_module.py
```diff
@@ -1,11 +1,19 @@
...
+logger = logging.getLogger(__name__)
+
 
 async def request_raw(request: fastapi.Request) -> dict:
-    return await request.json()
+    try:
+        return await request.json()
+    except json.JSONDecodeError as e:
+        logger.warning(f"Unable to get JSON from request. Error is:\n{e}")
+        return {}
```
elijahbenizzy, line 16: Why shouldn't this be a 4XX?

skrawcz: the json body is empty. So it's not a bad request.

elijahbenizzy (approved):

PR #217, opened 2022-10-27
Example showing what we would need to modify to enable user question

skrawcz:
If the user wants to reuse hamilton functions without hamilton, this commit shows one way to do it -- and what would be required on Hamilton's side to support it.

Basically we'd need a convention. Not sure it's a good idea. Probably easier to instead have boilerplate code people use to try/except the import if hamilton does not exist and have the decorators be identity functions...

### DIFF
--- a/examples/user_help/module_import_question/README.md
+++ b/examples/user_help/module_import_question/README.md
@@ -1,0 +1,3 @@
+Context:
+ - hamilton is required as a dependency in the python module if you use any decorators
+ - how can we reuse the functions without having to have hamilton?
--- a/examples/user_help/module_import_question/hamilton_plain_functions.py
+++ b/examples/user_help/module_import_question/hamilton_plain_functions.py
@@ -1,0 +1,15 @@
+"""Module that brings in Hamilton as a dependency"""
+
+import plain_functions
+
+from hamilton import function_modifiers
+
+moving_average__v1 = function_modifiers.config.when(ma_type="v1")(
+    plain_functions.moving_average__v1
+)
+
+moving_average__v2 = function_modifiers.config.when(ma_type="v2")(
+    plain_functions.moving_average__v2
+)
+
+some_other_function = plain_functions.some_other_function
--- a/examples/user_help/module_import_question/plain_functions.py
+++ b/examples/user_help/module_import_question/plain_functions.py
@@ -1,0 +1,16 @@
+"""Module that defines functions without importing Hamilton"""
+
+
+def moving_average__v1(input: dict) -> float:
+    # some logic
+    return 2.0
+
+
+def moving_average__v2(input: dict, other_input: float) -> float:
+    # some logic
+    return 2.0 * other_input
+
+
+def some_other_function(moving_average: float) -> float:
+    # some logic
+    return moving_average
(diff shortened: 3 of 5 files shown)


elijahbenizzy:
I like my (somewhat hacky) approach more -- what you suggested earlier. This is easy, can be copy/pasted anywhere, and doesn't require a framework change:

```python
class mock_modifier:
    def __getattr__(self, attr):
        return self

    def __call__(*args, **kwargs):
        def identity(fn):
            return fn
        return identity

try:
    from hamilton import function_modifiers
except ImportError as e:
    print('Cannot import hamilton, using a mock function modifier')
    function_modifiers = mock_modifier()


@function_modifiers.config.when(foo=1)
def bar() -> int:
    return 1

@function_modifiers.extract_fields({'a' : int, 'b' : int})
def baz() -> dict:
    return {'a' : 1, 'b' : 1}


if __name__ == '__main__':
    print(bar())
    print(baz())
```

skrawcz:
another idea I had, was to have a dummy python module that one could import e.g. sf-hamilton-noop ...

skrawcz:
> I like my (somewhat hacky) approach more -- what you suggested above. This is easy, can be copy/pasted anywhere, and doesn't require a framework change:
> 
> ```python
> class mock_modifier:
>     def __getattr__(self, attr):
>         return self
> 
>     def __call__(*args, **kwargs):
>         def identity(fn):
>             return fn
>         return identity
> 
> try:
>     from hamilton import function_modifiers
> except ImportError as e:
>     print('Cannot import hamilton, using a mock function modifier')
>     function_modifiers = mock_modifier()
> 
> 
> @function_modifiers.config.when(foo=1)
> def bar() -> int:
>     return 1
> 
> @function_modifiers.extract_fields({'a' : int, 'b' : int})
> def baz() -> dict:
>     return {'a' : 1, 'b' : 1}
> 
> 
> if __name__ == '__main__':
>     print(bar())
>     print(baz())
> ```

would this work for the case someone imports from the module? It wouldn't right?
E.g. 
```python
from hamilton.function_modifiers import config
```

elijahbenizzy:
> > I like my (somewhat hacky) approach more -- what you suggested above. This is easy, can be copy/pasted anywhere, and doesn't require a framework change:
> > ```python
> > class mock_modifier:
> >     def __getattr__(self, attr):
> >         return self
> > 
> >     def __call__(*args, **kwargs):
> >         def identity(fn):
> >             return fn
> >         return identity
> > 
> > try:
> >     from hamilton import function_modifiers
> > except ImportError as e:
> >     print('Cannot import hamilton, using a mock function modifier')
> >     function_modifiers = mock_modifier()
> > 
> > 
> > @function_modifiers.config.when(foo=1)
> > def bar() -> int:
> >     return 1
> > 
> > @function_modifiers.extract_fields({'a' : int, 'b' : int})
> > def baz() -> dict:
> >     return {'a' : 1, 'b' : 1}
> > 
> > 
> > if __name__ == '__main__':
> >     print(bar())
> >     print(baz())
> > ```
> 
> would this work for the case someone imports from the module? It wouldn't right? E.g.
> 
> ```python
> from hamilton.function_modifiers import config
> ```

Yeah it would have to be adjusted...

PR #233, opened 2022-11-17
Running Hamilton on Flyte

ramannanda9:

## Checklist

- [X] PR has an informative and human-readable title (this will be pulled into the release notes)
- [X] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.


hamilton/experimental/h_flytekit.py
```diff
@@ -0,0 +1,141 @@
...
+        # define a python task
+        task = flytekit.PythonFunctionTask(task_config=None, task_function=node.callable)
+        input_kwargs = {}
+        # map inputs
+        for input_node in node.dependencies:
+            if input_node.name not in self.workflow.inputs:
+                self.workflow.add_workflow_input(input_node.name, input_node.type)
+            input_kwargs[input_node.name] = self.workflow.inputs[input_node.name]
+        # add the node to workflow
+        wf_node = self.workflow.add_entity(task, **input_kwargs)
```
elijahbenizzy, line 119: I might be misreading, but why would it already have outputs/where does it get it from?

hamilton/experimental/h_flytekit.py
```diff
@@ -0,0 +1,141 @@
...
+        if wf_node.outputs:
+            outputs = {}
+            for output_name, output_promise in wf_node.outputs.items():
+                outputs[output_name] = output_promise
+            # handle single output case
+            if len(outputs) == 1:
+                self.workflow.add_workflow_output(node.name, list(outputs.values())[0], node.type)
+            else:
+                self.workflow.add_workflow_output(node.name, outputs, node.type)
```
elijahbenizzy, line 128: And why might it have multiple outputs?

elijahbenizzy (commented):
Some thoughts -- don't fully understand it -- I need to look into the flytekit API. I think we might want a better toolset for compiling to other frameworks...

hamilton/experimental/h_flytekit.py
```diff
@@ -0,0 +1,141 @@
...
+
+class FlyteKitAdapter(base.HamiltonGraphAdapter, base.ResultMixin):
+    """Class representing what's require to run Hamilton on Flytekit
+
+    Use `pip install sf-hamilton[flytekit] to get the dependencies required to run this.
+
+    Flytekit Python is the Python Library for easily authoring, testing, deploying,
+    and interacting with Flyte tasks, workflows, and launch plans
+
+    """
```
skrawcz, line 84: On the right track @ramannanda9 !

Here we'd probably need an explanation here of how we're actually using Flyte.

hamilton/experimental/h_flytekit.py
```diff
@@ -0,0 +1,141 @@
...
+        self, ctx: FlyteContext, python_val: pd.Series, expected_python_type: Type[pd.Series]
+    ) -> str:
+        return python_val.describe().to_string()
+
+    def _get_schema_type(self):
+        return SchemaType(columns=[])
+
+
+# Register with FlyteKit type engine
+TypeEngine.register(PandasSeriesTransformer())
```
skrawcz, line 73: @ramannanda9 since Hamilton supports any object type, we need to think how to handle arbitrary python object types here too.

ramannanda9: Yeah not the best idea, but they ideally could default to just pickle dumping rather than raising unsupported type.

hamilton/experimental/h_flytekit.py
ramannanda9: Yep, See [Imperative Workflow](https://docs.flyte.org/projects/cookbook/en/stable/auto/core/flyte_basics/imperative_wf_style.html ) . It happens when you add the entity i.e. task with `input_kwargs`

hamilton/experimental/h_flytekit.py
```diff
@@ -0,0 +1,141 @@
...
+        :param kwargs: the arguments that should be passed to it
+        :return: Create a flytekit task and execute it?
+        """
+        # define a python task
+        task = flytekit.PythonFunctionTask(task_config=None, task_function=node.callable)
+        input_kwargs = {}
+        # map inputs
+        for input_node in node.dependencies:
+            if input_node.name not in self.workflow.inputs:
+                self.workflow.add_workflow_input(input_node.name, input_node.type)
```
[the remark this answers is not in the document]
ramannanda9: There is fairly limited support for custom types, but yeah a check can be added.

PR #243, opened 2022-12-12
PandasDataFrameResult: Convert non-list values to single row frame

ianhoffman:
When trying to run an intermediate node which produces a scalar in the `hello_world` example, Hamilton throws an error:

```
WARNING:hamilton.base:It appears no Pandas index type was detected. This will likely break when trying to create a DataFrame. E.g. are you requesting all scalar values? Use a different result builder or return at least one Pandas object with an index. Ignore this warning if you're using DASK for now.
ERROR:hamilton.driver:-------------------------------------------------------------------
Oh no an error! Need help with Hamilton?
Join our slack and ask for help! https://join.slack.com/t/hamilton-opensource/shared_invite/zt-1bjs72asx-wcUTgH7q7QX1igiQ5bbdcg
-------------------------------------------------------------------

Traceback (most recent call last):
  File "my_script.py", line 29, in <module>
    df = dr.execute(output_columns)
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/hamilton/driver.py", line 142, in execute
    raise e
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/hamilton/driver.py", line 139, in execute
    return self.adapter.build_result(**outputs)
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/hamilton/base.py", line 171, in build_result
    raise ValueError(f"Cannot build result. Cannot handle type {value}.")
ValueError: Cannot build result. Cannot handle type 28.333333333333332.
```

If we can run an entire DAG, it seems like we should be able to run any sub-DAG of the DAG.

## Changes

Updates `PandasDataFrameResult.build_result()` to convert scalar values into dataframes.

## How I tested this

Updated unit tests.

## Checklist

- [X] PR has an informative and human-readable title (this will be pulled into the release notes)
- [X] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future TODOs are captured in comments
- [X] Project documentation has been updated if adding/changing functionality.

### DIFF
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -166,12 +166,10 @@ class PandasDataFrameResult(ResultMixin):
             (value,) = outputs.values()  # this works because it's length 1.
             if isinstance(value, pd.DataFrame):
                 return value
-            elif isinstance(value, pd.Series):
-                return pd.DataFrame(outputs)
 
         if not any(pd.api.types.is_list_like(value) for value in outputs.values()):
-            # If we're dealing with all scalar values,
-            # coerce the output to a single-row, multi-column dataframe.
+            # If we're dealing with all values that don't have any "index" that could be created
+            # (i.e. scalars, objects) coerce the output to a single-row, multi-column dataframe.
             return pd.DataFrame([outputs])
 
         return pd.DataFrame(outputs)
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -169,7 +169,7 @@ class PandasDataFrameResult(ResultMixin):
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
 
-        if not any(isinstance(value, (pd.DataFrame, pd.Series)) for value in outputs.values()):
+        if not any(pd.api.types.is_list_like(value) for value in outputs.values()):
             # If we're dealing with all scalar values,
             # coerce the output to a single-row, multi-column dataframe.
             return pd.DataFrame([outputs])
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -168,7 +168,8 @@ class PandasDataFrameResult(ResultMixin):
                 return value
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
-            raise ValueError(f"Cannot build result. Cannot handle type {value}.")
+            else:
+                return pd.DataFrame([outputs])
         return pd.DataFrame(outputs)
 
 
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -168,8 +168,12 @@ class PandasDataFrameResult(ResultMixin):
                 return value
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
-            else:
-                return pd.DataFrame([outputs])
+
+        if not any(isinstance(value, (pd.DataFrame, pd.Series)) for value in outputs.values()):
+            # If we're dealing with all scalar values,
+            # coerce the output to a single-row, multi-column dataframe.
+            return pd.DataFrame([outputs])
+
         return pd.DataFrame(outputs)
 
 
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -235,7 +235,6 @@ def test_PandasDataFrameResult_build_result(outputs, expected_result):
         ({"a": [1, 2], "b": {"foo": "bar"}}),
         ({"a": [1, 2], "b": [3, 4, 5]}),
         ({"a": np.array([1, 2]), "b": np.array([3, 4, 5])}),
-        ({"a": np.array([1, 2]), "b": np.array([[3, 4], [5, 6]])}),
         ({"a": _gen_ints(3), "b": _gen_ints(4)}),
     ],
     ids=[
@@ -244,7 +243,6 @@ def test_PandasDataFrameResult_build_result(outputs, expected_result):
         "test-lists-and-dicts",
         "test-mismatched-lists",
         "test-mismatched-arrays",
-        "test-mismatched-dicts",
         "test-mismatched-generators",
     ],
 )
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -147,6 +147,7 @@ def test_SimplePythonDataFrameGraphAdapter_check_input_type_mismatch(node_type, 
                 {"a": pd.Series([1, 2, 3]), "b": pd.Series([11, 12, 13]), "c": pd.Series([0, 1, 2])}
             ),
         ),
+        ({"a": 1}, pd.DataFrame([{"a": 1}])),
     ],
     ids=[
         "test-single-series",
@@ -154,6 +155,7 @@ def test_SimplePythonDataFrameGraphAdapter_check_input_type_mismatch(node_type, 
         "test-multiple-series",
         "test-multiple-series-with-scalar",
         "test-multiple-series-with-index",
+        "test-single-value",
     ],
 )
 def test_PandasDataFrameResult_build_result(outputs, expected_result):
@@ -166,7 +168,6 @@ def test_PandasDataFrameResult_build_result(outputs, expected_result):
 @pytest.mark.parametrize(
     "outputs",
     [
-        ({"a": 1}),
         (
             {
                 "a": pd.DataFrame({"a": [1, 2, 3], "b": [11, 12, 13]}),
@@ -182,7 +183,6 @@ def test_PandasDataFrameResult_build_result(outputs, expected_result):
         ),
     ],
     ids=[
-        "test-single-value",
         "test-multiple-dataframes",
         "test-multiple-series-with-dataframe",
     ],
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -122,11 +122,13 @@ def test_SimplePythonDataFrameGraphAdapter_check_input_type_mismatch(node_type, 
 @pytest.mark.parametrize(
     "outputs,expected_result",
     [
+        ({"a": 1}, pd.DataFrame([{"a": 1}])),
         ({"a": pd.Series([1, 2, 3])}, pd.DataFrame({"a": pd.Series([1, 2, 3])})),
         (
             {"a": pd.DataFrame({"a": [1, 2, 3], "b": [11, 12, 13]})},
             pd.DataFrame({"a": pd.Series([1, 2, 3]), "b": pd.Series([11, 12, 13])}),
         ),
+        ({"a": 1, "bar": 2}, pd.DataFrame([{"a": 1, "bar": 2}])),
         (
             {"a": pd.Series([1, 2, 3]), "b": pd.Series([11, 12, 13])},
             pd.DataFrame({"a": pd.Series([1, 2, 3]), "b": pd.Series([11, 12, 13])}),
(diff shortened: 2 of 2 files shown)


skrawcz:
Awesome thanks.

Note: This will still break if you try to get multiple "scalar" values back. 🤔 
But I think we have enough information to do the same fix with the proper if statement.

ianhoffman:
> Awesome thanks.
> 
> Note: This will still break if you try to get multiple "scalar" values back. 🤔 But I think we have enough information to do the same fix with a "requisite" if statement.
> 
> It's getting late, but we can chat and see if you'd like to add these extras, or I can do it.

Pushed a fix... let me know what you think. I guess I also could have done this by checking whether `len(all_index_types) == len(no_indexes)`. But that felt like maybe not the intended usage of `PandasDataFrameResult.pandas_index_types`.

hamilton/base.py
```diff
@@ -168,7 +168,12 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
                 return value
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
-            raise ValueError(f"Cannot build result. Cannot handle type {value}.")
+
+        if not any(isinstance(value, (pd.DataFrame, pd.Series)) for value in outputs.values()):
```
ianhoffman, line 172: We could also check for scalars explicitly here, e.g. `(int, str, float, bool)` or something like that. Since this will break for valid outputs:

```python
>>> pd.DataFrame({'foo': [1, 2], 'bar': [3, 4]})
   foo  bar
0    1    3
1    2    4
>>> pd.DataFrame([{'foo': [1, 2], 'bar': [3, 4]}])
      foo     bar
0  [1, 2]  [3, 4]
```

Don't know how much we want to worry about this edge case, though.

elijahbenizzy: Could specifically check for list, but I think this is OK. My opinion is that the goal of this should be (a) to get it not to break and (b) to get it not to lose any information. If folks don't like the way we join it they can use the dict result or build a custom `ResultBuilder`. Might be nice logging a warning though with the info so they know what to do...

skrawcz:
Actually it seems we should be able rely more on pandas here -- e.g. https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_list_like.html 🤔

hamilton/base.py
ianhoffman: Hmm — we already do log some sort of warning:

```
❯ python3 my_script.py
WARNING:hamilton.base:It appears no Pandas index type was detected. This will likely break when trying to create a DataFrame. E.g. are you requesting all scalar values? Use a different result builder or return at least one Pandas object with an index. Ignore this warning if you're using DASK for now.
   spend_std_dev
0      17.224014
```

This isn't the most intuitive though, since things now work when you request all scalar values (but may not work when you request a list).

@skrawcz said he was gonna revisit that error message; my two cents are that we could

* log nothing if all scalars are requested (since it works)
* log if lists are requested, OR do a best effort attempt to resolve lists into dataframes. Probably logging is good enough.

skrawcz: the general case we're dealing with is if there is no "index" for pandas to know what to do.

Test cases to cover:

* list like things (e.g. builtin types and np types and "sequence"/"generator" types)
* scalars
* other objects

and then the mixture of them, e.g. list & scalar == error,  list & list == fine (if they're the same length), scalar & scalar == fine, scalar & object == fine, etc.

ianhoffman: I think that we only need to worry about cases where `pd.DataFrame(outputs)` won't work... e.g. when we need to do `pd.DataFrame([outputs])` instead. Yeah, there are times when `pd.DataFrame(outputs)` will blow up – e.g. if you pass a combination of dicts and lists — which we aren't yet validating, but I think that's out of scope for now (not that we should never take it on). So I'm tempted to say that the code is good enough as is (pending a switch to `is_list_like`) and that we should add tests for other cases but not explicitly handle them here.

ianhoffman: E.g. pandas handles these cases perfectly fine:

```python
>>> pd.DataFrame({'foo': {'a': 1, 'b': 2}, 'bar': {'c': 2, 'd': 3}})
   foo  bar
a  1.0  NaN
b  2.0  NaN
c  NaN  2.0
d  NaN  3.0
>>> pd.DataFrame({'foo': [3, 4], 'bar': [1, 2]})
   foo  bar
0    3    1
1    4    2
```

And it blows up in an informative way here:

```python
>>> pd.DataFrame({'foo': [3, 4], 'bar': {'a': 1, 'b': 2}})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/pandas/core/frame.py", line 663, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 493, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 118, in arrays_to_mgr
    index = _extract_index(arrays)
  File "/Users/ian.hoffman/src/hamilton/examples/hello_world/.venv/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 669, in _extract_index
    raise ValueError(
ValueError: Mixing dicts with non-Series may lead to ambiguous ordering.
```

Casting to a list does fix the above error, but the output isn't worth much:

```python
>>> pd.DataFrame([{'foo': [3, 4], 'bar': {'a': 1, 'b': 2}}])
      foo               bar
0  [3, 4]  {'a': 1, 'b': 2}
```

elijahbenizzy: Yeah I'm wondering if this set of rules is just too complicated and it should break when there's no index. Might not be worth trying to intuit it...

So:

1 series/DF or more -> works
0 series -> breaks

Any more complex and we're setting ourselves up for unintuitive behavior later. Maybe the break says "You should use the dict adapter if you don't have a dataframe output"

ianhoffman: If we did that, we'd need to swap out adapters to run intermediate nodes, which seems painful for debugging. I think it's pretty reasonable to convert scalars into a single-row, multi-column DF and display them — that's helpful for end-users and not super complex.

I do think anything beyond that is overdoing it, though.

hamilton/base.py
```diff
@@ -168,7 +168,12 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
                 return value
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
-            raise ValueError(f"Cannot build result. Cannot handle type {value}.")
+
+        if not any(pd.api.types.is_list_like(value) for value in outputs.values()):
```
skrawcz, line 172: I think we need to add `pd.api.types.is_dict_like(value)` too.

skrawcz: so pd.DataFrame({'a': {'x': 1, 'y':2}, 'z': 12}) works.

hamilton/base.py
```diff
@@ -168,7 +168,12 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
                 return value
             elif isinstance(value, pd.Series):
                 return pd.DataFrame(outputs)
```
skrawcz, line 170: with your code below, I believe we can kill these two lines as well.

hamilton/base.py
```diff
@@ -168,7 +168,12 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
...
-            raise ValueError(f"Cannot build result. Cannot handle type {value}.")
+
+        if not any(pd.api.types.is_list_like(value) for value in outputs.values()):
+            # If we're dealing with all scalar values,
```
skrawcz, line 173:
```suggestion
            # If we're dealing with all values that don't have any "index" that could be created (i.e. scalars, objects)
```

hamilton/base.py
skrawcz, line 170:
```suggestion
```

hamilton/base.py
skrawcz:
```suggestion
        if not any(pd.api.types.is_list_like(value) or pd.api.types.is_dict_like(value) for value in outputs.values()):
```

tests/test_base.py
```diff
@@ -147,13 +159,54 @@ def test_SimplePythonDataFrameGraphAdapter_check_input_type_mismatch(node_type,
...
         "test-multiple-series-with-scalar",
         "test-multiple-series-with-index",
+        "test-multiple-lists",
+        "test-multiple-arrays",
+        "test-multiple-dicts",
+        "test-multiple-generators",
+        "test-scalar-and-series",
+        "test-scalar-and-list",
+        "test-series-and-list",
     ],
```
skrawcz, line 210: just need to add:
 - test-single-object
 - tests-dict-and-scalar

hamilton/base.py
skrawcz: actually we don't -- dict resolves as "list like".

ianhoffman: `is_list_like` seems to work on dicts too:

```python
>>> pd.api.types.is_list_like({'a': {'x': 1, 'y':2}, 'z': 12})
True
```

Still want this for explicit-ness/futureproofing?

hamilton/base.py
ianhoffman: Good catch!

tests/test_base.py
```diff
@@ -119,14 +119,26 @@ def test_SimplePythonDataFrameGraphAdapter_check_input_type_mismatch(node_type,
     assert actual is False
 
 
+def _gen_ints(n: int) -> typing.Iterator[int]:
```
elijahbenizzy, line 122: Very nitty nit pick:
1. typing could be `Generator` 
2. You can use `yield from` 
```python
def _gen_int(n) -> Generator[int, None, None]:
   ...:     yield from range(n)
```

skrawcz (approved):
I think this is good enough!

Great thanks for the help @ianhoffman !

skrawcz:
Only TODO will be to update the check index warning.

ianhoffman:
> Only TODO will be to update the check index warning.

Yeah, I can do that too - just not sure what we want the behavior to be. Maybe we just shouldn’t log if all scalars/objects were provided? Since we are providing this IMO pretty reasonable output in that case. Or did you have something else in mind?

skrawcz:
> > Only TODO will be to update the check index warning.
> 
> Yeah, I can do that too - just not sure what we want the behavior to be. Maybe we just shouldn’t log if all scalars/objects were provided? Since we are providing this IMO pretty reasonable output in that case. Or did you have something else in mind?

We can do that in a follow up PR/commit.

skrawcz:
@ianhoffman https://github.com/stitchfix/hamilton/pull/246

hamilton/base.py
ianhoffman: Missed this.

> this is for the use-case of just running a piece of the DAG, right?

Yup.

> I wonder if there's a special way to provide a spine. E.G. have something like a configuration element that the DAG-builder knows about, or something similar. Also we could have guess_spine as a parameter to enable less strict checking...

That could work. Alternatively my kernel could just provide its own custom result builder which does the behavior implemented in this PR.

I do think most users won't care about these sort of advanced configuration options though; they probably expect Hamilton to "just work" without any configuration. It seems to me that visualizing >=1 scalar values is a pretty legit use-case, so IMO that should "just work" without config. Whether we allow config to support more advanced use cases is a different question. I do think we should support config for advanced use cases, but I don't think it's urgent.

Issue #244, opened 2022-12-14
Cannot use @does with @extract_columns


skrawcz:
Cannot use `@does` with `@extract_columns` when used together on a function.

# Current behavior

The code fails on DAG construction with a function defined like so:

```python
def _cross_join(**series: pd.Series) -> pd.DataFrame:
   # logic...

@extract_columns('a', 'b')
@does(_cross_join)
def field1_field2_cross_join(sample_field1: pd.Series, sample_field2: pd.Series) -> pd.DataFrame:
    pass
```

## Stack Traces

```python
Traceback (most recent call last):
  File "/Users/stefankrawczyk/dw/user_help/does_bug/run.py", line 9, in <module>
    dr = driver.Driver({}, functions)
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/driver.py", line 62, in __init__
    raise e
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/driver.py", line 59, in __init__
    self.graph = graph.FunctionGraph(*modules, config=config, adapter=adapter)
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/graph.py", line 184, in __init__
    self.nodes = create_function_graph(*modules, config=self._config, adapter=adapter)
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/graph.py", line 92, in create_function_graph
    for n in function_modifiers_base.resolve_nodes(f, config):
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/function_modifiers_base.py", line 356, in resolve_nodes
    nodes = node_expander.transform_dag(nodes, config, fn)
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/function_modifiers_base.py", line 162, in transform_dag
    return self.expand_node(node_, config, fn)
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/function_modifiers.py", line 444, in expand_node
    node.Node(
  File "/Users/stefankrawczyk/dw/hamilton/hamilton/node.py", line 87, in __init__
    raise ValueError(
ValueError: Missing type hint for _does__fn in function field1_field2_cross_join. Please add one to fix.
```

## Steps to replicate behavior
```python
# functions.py
import pandas as pd
from hamilton.function_modifiers import does, extract_columns

def _cross_join(**series: pd.Series) -> pd.DataFrame:
    return pd.DataFrame({'a': [1, 2, 3], 'b': [3, 4, 5]})

[...]

PR #245, opened 2022-12-14
Fixes does function wrapping #244

skrawcz:
In issue #244 types were not being propagated of the function being wrapped.

This changes that by adding the functools.wraps decorator.

Adds a unit test to test for two decorators playing nice together and verified that it's broken before and fixed afterwards.

## Changes
- modifies `@does` wrapper function to properly emulate the signature of the function it is replacing.

## How I tested this
- locally
- unit test

## Checklist

### DIFF
--- a/hamilton/function_modifiers.py
+++ b/hamilton/function_modifiers.py
@@ -744,7 +744,8 @@ class does(function_modifiers_base.NodeCreator):
         and the same parameters/types as the original function.
         """
 
-        def replacing_function(__fn=fn, **kwargs):
+        @functools.wraps(fn)
+        def wrapper_function(**kwargs):
             final_kwarg_values = {
                 key: param_spec.default
                 for key, param_spec in inspect.signature(fn).parameters.items()
@@ -754,7 +755,7 @@ class does(function_modifiers_base.NodeCreator):
             final_kwarg_values = does.map_kwargs(final_kwarg_values, self.argument_mapping)
             return self.replacing_function(**final_kwarg_values)
 
-        return node.Node.from_fn(fn).copy_with(callabl=replacing_function)
+        return node.Node.from_fn(fn).copy_with(callabl=wrapper_function)
 
 
 class dynamic_transform(function_modifiers_base.NodeCreator):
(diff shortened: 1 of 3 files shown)


hamilton/function_modifiers.py
```diff
@@ -754,7 +755,7 @@ def replacing_function(__fn=fn, **kwargs):
             final_kwarg_values = does.map_kwargs(final_kwarg_values, self.argument_mapping)
             return self.replacing_function(**final_kwarg_values)
 
-        return node.Node.from_fn(fn).copy_with(callabl=replacing_function)
+        return node.Node.from_fn(fn).copy_with(callabl=wrapper_function)
```
elijahbenizzy, line 758: OK, so let's ship this as it fixes it. I'm confused as to how this didn't take the original function: https://github.com/stitchfix/hamilton/blob/37317b19de79f6b24ae08c0f478fd62e3646cbf6/hamilton/node.py#L201. It should have used the initial type... Need to dig in.

tests/test_graph.py
```diff
@@ -359,6 +360,20 @@ def test_end_to_end_with_column_extractor_nodes():
...
 
 
+def test_end_to_end_with_multiple_decorators():
+    """Tests that a simple function graph with multiple decorators on a function works end-to-end"""
+    fg = graph.FunctionGraph(
+        tests.resources.multiple_decorators_together, config={"param0": 3, "param1": 1}
+    )
+    nodes = fg.get_nodes()
+    results = fg.execute(nodes, {}, {})
+    print(results)
```
elijahbenizzy, line 370: printline

elijahbenizzy (approved):
Left one comment about something that doesn't make sense but let's ship for the users.

hamilton/function_modifiers.py
skrawcz: No no, that's not the fix.

1. I found it confusing that the callable created was named the same as the internal function.
2. The actual fix is the functools.wraps and removing `__fn` parameter.

skrawcz:
Closing in favor of #247 .

PR #246, opened 2022-12-14
Adjusts index type check warnings

skrawcz:
Following #243 we need to adjust the warning
to make sense. Since we are not in danger of things breaking imminently. So the warning is just there for the user now to indicate how Pandas might be resolving indexes to create the dataframe.

## Changes
* adjusts logger warning message to now match behavior after #243 .

## How I tested this
* locally.

## Checklist


skrawcz:
@ianhoffman I think this should do the trick.

PR #255, opened 2022-12-21
Adds telemetry

skrawcz:
This adds telemetry capture to Hamilton.

After this change, by default, when using Hamilton, it will collect anonymous usage data to help us improve Hamilton and know where to apply development efforts.

We capture two events: one when a driver object is instantiated, and one when the `execute()` call on the driver completes.
No user data or potentially sensitive information is or ever will be collected. The captured data is limited to:

* Operating System and Python version
* A persistent UUID to indentify the session, stored in ~/.hamilton.conf. 
* Error stack trace limited to Hamilton code, if one occurs.
* Information on what features you're using from Hamilton: decorators, adapters, result builders.
* How Hamilton is being used: number of final nodes in DAG, number of modules, size of objects passed to `execute()`. 

If you do not wish to participate, one can opt-out with one of the following methods:
1. Set it to false programmatically in your code before creating a Hamilton driver:
   ```python
   from hamilton import telemetry
   telemetry.disable_telemetry()
   ```
2. Set the key `telemetry_enabled` to `false` in ~/.hamilton.conf under the `DEFAULT` section:
   ```
   [DEFAULT]
   telemetry_enabled = True
   ```
3. Set HAMILTON_TELEMETRY_ENABLED=false as an environment variable. Either setting it for your shell session:
   ```bash
   export HAMILTON_TELEMETRY_ENABLED=false
   ```
   or passing it as part of the run command:
   ```bash
   HAMILTON_TELEMETRY_ENABLED=false python NAME_OF_MY_DRIVER.py 
   ```

-------
## Changes
 - adds telemetry.py
 - adds code to driver.Driver to capture telemetry
 - the code should be wrapped in enough try excepts that it should not impact running operations
 - adds code to function_modifiers base.
 - adds tests 

## How I tested this
Tested this locally, and added unit tests.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,41 @@ If you are doing a lot of repetitive work, one might consider multiple cursors. 
 
 To use it hit `option + mouse click` to create multiple cursors. `Esc` to revert back to a normal mode.
 
+# Usage analytics & data privacy
+By default, when using Hamilton, it collects anonymous usage data to help improve Hamilton and know where to apply development
+efforts.
+
+We capture two events: one when the `Driver` object is instantiated, and one when the `execute()` call on the `Driver` object completes.
+No user data or potentially sensitive information is or ever will be collected. The captured data is limited to:
+
+* Operating System and Python version
+* A persistent UUID to indentify the session, stored in ~/.hamilton.conf.
+* Error stack trace limited to Hamilton code, if one occurs.
+* Information on what features you're using from Hamilton: decorators, adapters, result builders.
+* How Hamilton is being used: number of final nodes in DAG, number of modules, size of objects passed to `execute()`.
+
+If you're worried, see telemetry.py for details.
+
+If you do not wish to participate, one can opt-out with one of the following methods:
+1. Set it to false programmatically in your code before creating a Hamilton driver:
+   ```python
+   from hamilton import telemetry
+   telemetry.disable_telemetry()
+   ```
+2. Set the key `telemetry_enabled` to `false` in ~/.hamilton.conf under the `DEFAULT` section:
+   ```
+   [DEFAULT]
+   telemetry_enabled = True
+   ```
+3. Set HAMILTON_TELEMETRY_ENABLED=false as an environment variable. Either setting it for your shell session:
+   ```bash
+   export HAMILTON_TELEMETRY_ENABLED=false
+   ```
+   or passing it as part of the run command:
+   ```bash
+   HAMILTON_TELEMETRY_ENABLED=false python NAME_OF_MY_DRIVER.py
+   ```
+
 # Contributors
 
 ## Code Contributors
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ No user data or potentially sensitive information is or ever will be collected. 
 * A persistent UUID to indentify the session, stored in ~/.hamilton.conf.
 * Error stack trace limited to Hamilton code, if one occurs.
 * Information on what features you're using from Hamilton: decorators, adapters, result builders.
-* How Hamilton is being used: number of final nodes in DAG, number of modules, size of objects passed to `execute()`.
+* How Hamilton is being used: number of final nodes in DAG, number of modules, size of objects passed to `execute()`, the name of the Driver function being invoked.
 
 If you're worried, see telemetry.py for details.
 
--- a/examples/async/fastapi_example.py
+++ b/examples/async/fastapi_example.py
@@ -1,17 +1,23 @@
 import async_module
 import fastapi
 
+from hamilton import base
 from hamilton.experimental import h_async
 
 app = fastapi.FastAPI()
+
+# can instantiate a driver once for the life of the app:
+dr = h_async.AsyncDriver({}, async_module, result_builder=base.DictResult())
 
 
 @app.post("/execute")
 async def call(request: fastapi.Request) -> dict:
     """Handler for pipeline call"""
-    dr = h_async.AsyncDriver({}, async_module)
     input_data = {"request": request}
-    return await dr.raw_execute(["pipeline"], inputs=input_data)
+    # Can instantiate a driver within a request as well:
+    # dr = h_async.AsyncDriver({}, async_module, result_builder=base.DictResult())
+    result = await dr.execute(["pipeline"], inputs=input_data)
+    return result
 
 
 if __name__ == "__main__":
--- a/graph_adapter_tests/h_async/test_h_async.py
+++ b/graph_adapter_tests/h_async/test_h_async.py
@@ -66,7 +66,9 @@ async def test_driver_end_to_end():
 @mock.patch("hamilton.telemetry.g_telemetry_enabled", True)
 async def test_driver_end_to_end_telemetry(send_event_json):
     dr = h_async.AsyncDriver({}, simple_async_module, result_builder=base.DictResult())
-    all_vars = [var.name for var in dr.list_available_variables()]
+    with mock.patch("hamilton.telemetry.g_telemetry_enabled", False):
+        # don't count this telemetry tracking invocation
+        all_vars = [var.name for var in dr.list_available_variables()]
     result = await dr.execute(final_vars=all_vars, inputs={"external_input": 1})
     assert result == {
         "another_async_func": 8,
--- a/graph_adapter_tests/h_async/test_h_async.py
+++ b/graph_adapter_tests/h_async/test_h_async.py
@@ -1,7 +1,9 @@
 import asyncio
+from unittest import mock
 
 import pytest
 
+from hamilton import base
 from hamilton.experimental import h_async
 
 from .resources import simple_async_module
@@ -57,3 +59,25 @@ async def test_driver_end_to_end():
         "simple_async_func": 2,
         "simple_non_async_func": 7,
     }
+
+
+@pytest.mark.asyncio
+@mock.patch("hamilton.telemetry.send_event_json")
+@mock.patch("hamilton.telemetry.g_telemetry_enabled", True)
+async def test_driver_end_to_end_telemetry(send_event_json):
+    dr = h_async.AsyncDriver({}, simple_async_module, result_builder=base.DictResult())
+    all_vars = [var.name for var in dr.list_available_variables()]
+    result = await dr.execute(final_vars=all_vars, inputs={"external_input": 1})
+    assert result == {
+        "another_async_func": 8,
+        "async_func_with_param": 4,
+        "external_input": 1,
+        "non_async_func_with_decorator": {"result_1": 9, "result_2": 5},
+        "result_1": 9,
+        "result_2": 5,
+        "simple_async_func": 2,
+        "simple_non_async_func": 7,
+    }
+    await asyncio.sleep(1)  # to ensure the last telemetry invocation finishes executing
+    assert send_event_json.called
+    assert len(send_event_json.call_args_list) == 2
--- a/graph_adapter_tests/h_async/test_h_async.py
+++ b/graph_adapter_tests/h_async/test_h_async.py
@@ -78,6 +78,15 @@ async def test_driver_end_to_end_telemetry(send_event_json):
         "simple_async_func": 2,
         "simple_non_async_func": 7,
     }
-    await asyncio.sleep(1)  # to ensure the last telemetry invocation finishes executing
+    # to ensure the last telemetry invocation finishes executing
+    # get all tasks -- and the current task, and await all others.
+    try:
+        # only works for 3.7+
+        tasks = asyncio.all_tasks()
+        current_task = asyncio.current_task()
+        await asyncio.gather(*[t for t in tasks if t != current_task])
+    except AttributeError:
+        # required for 3.6
+        await asyncio.sleep(1)
     assert send_event_json.called
     assert len(send_event_json.call_args_list) == 2
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -13,7 +13,10 @@ import pandas as pd
 import typing_inspect
 from pandas.core.indexes import extension as pd_extension
 
-from . import node
+try:
+    from . import node
+except ImportError:
+    import node
 
 logger = logging.getLogger(__name__)
 
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1,3 +1,4 @@
+import functools
 import logging
 import sys
 import time
@@ -8,7 +9,7 @@ import uuid
 from dataclasses import dataclass, field
 from datetime import datetime
 from types import ModuleType
-from typing import Any, Collection, Dict, List, Optional, Tuple
+from typing import Any, Callable, Collection, Dict, List, Optional, Tuple
 
 import pandas as pd
 
@@ -28,6 +29,36 @@ else:
     from . import base, graph, node, telemetry
 
 logger = logging.getLogger(__name__)
+
+
+def capture_function_usage(call_fn: Callable) -> Callable:
+    """Decorator to wrap some driver functions for telemetry capture.
+
+    We want to use this for non-constructor and non-execute functions.
+    We don't capture information about the arguments at this stage,
+    just the function name.
+
+    :param call_fn: the Driver function to capture.
+    :return: wrapped function.
+    """
+
+    @functools.wraps(call_fn)
+    def wrapped_fn(*args, **kwargs):
+        try:
+            return call_fn(*args, **kwargs)
+        finally:
+            if telemetry.is_telemetry_enabled():
+                try:
+                    function_name = call_fn.__name__
+                    event_json = telemetry.create_driver_function_invocation_event(function_name)
+                    telemetry.send_event_json(event_json)
+                except Exception as e:
+                    if logger.isEnabledFor(logging.DEBUG):
+                        logger.error(
+                            f"Failed to send telemetry for function usage. Encountered:{e}\n"
+                        )
+
+    return wrapped_fn
 
 
 @dataclass
@@ -278,6 +309,7 @@ class Driver(object):
         del memoized_computation  # trying to cleanup some memory
         return outputs
 
+    @capture_function_usage
     def list_available_variables(self) -> List[Variable]:
         """Returns available variables, i.e. outputs.
 
@@ -285,6 +317,7 @@ class Driver(object):
         """
         return [Variable(node.name, node.type, node.tags) for node in self.graph.get_nodes()]
 
+    @capture_function_usage
     def display_all_functions(
         self, output_file_path: str, render_kwargs: dict = None, graphviz_kwargs: dict = None
     ):
@@ -304,6 +337,7 @@ class Driver(object):
         except ImportError as e:
             logger.warning(f"Unable to import {e}", exc_info=True)
 
+    @capture_function_usage
     def visualize_execution(
         self,
         final_vars: List[str],
@@ -340,6 +374,7 @@ class Driver(object):
         except ImportError as e:
             logger.warning(f"Unable to import {e}", exc_info=True)
 
+    @capture_function_usage
     def has_cycles(self, final_vars: List[str]) -> bool:
         """Checks that the created graph does not have cycles.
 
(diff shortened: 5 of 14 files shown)


skrawcz:
Quick review notes:

- [ ] check for module to not capture custom decorators
- [ ] wire up posthog (create account on free tier)
- [ ] write code to sanitize error trace to only capture hamilton related code
- [ ] write unit tests
- [ ] integration tests

elshize:
I'm wondering if you wouldn't also want to generate a unique ID for each execution to tie start and end events together. It would probably simplify data analysis, but also help to find started but unfinished runs.

skrawcz:
> I'm wondering if you wouldn't also want to generate a unique ID for each execution to tie start and end events together. It would probably simplify data analysis, but also help to find started but unfinished runs.

Good idea! Can add it.

skrawcz:
@elshize would you mind taking this branch for a spin? 
`pip install git+ssh://git@github.com/stitchfix/hamilton.git@add_telemetry`

elshize:
@skrawcz I updated Hamilton version in my projet and ran my pipeline, worked fine, and got these messages:

```
DEBUG:root:Succeed in sending telemetry consisting of [b'{"api_key": "phc_mZg8bkn3yvMxqvZKRlMlxjekFU5DFDdcdAsijJ2EH5e", "event": "os_hamilton_run_start", "properties": {"os_type": "posix", "os_version": "Linux-6.0.12-200.fc36.x86_64-x86_64-with-glibc2.35", "python_version": "3.9.11/CPython", "distinct_id": "cda55e8f-785c-4948-86ec-8b3ad3630d64", "hamilton_version:": [1, 11, 1], "telemetry_version": "0.0.1", "number_of_nodes": 21, "number_of_modules": 1, "number_of_config_items": 5, "decorators_used": {"extract_columns": 2}, "graph_adapter_used": "hamilton.base.SimplePythonDataFrameGraphAdapter", "result_builder_used": "custom_builder", "driver_run_id": "965d5d90-9e37-421c-b670-297abe1beddc", "error": null}}'].
DEBUG:root:Succeed in sending telemetry consisting of [b'{"api_key": "phc_mZg8bkn3yvMxqvZKRlMlxjekFU5DFDdcdAsijJ2EH5e", "event": "os_hamilton_run_end", "properties": {"os_type": "posix", "os_version": "Linux-6.0.12-200.fc36.x86_64-x86_64-with-glibc2.35", "python_version": "3.9.11/CPython", "distinct_id": "cda55e8f-785c-4948-86ec-8b3ad3630d64", "hamilton_version:": [1, 11, 1], "telemetry_version": "0.0.1", "is_success": true, "runtime_seconds": 0.3053734302520752, "number_of_outputs": 9, "number_of_overrides": 0, "number_of_inputs": 0, "driver_run_id": "965d5d90-9e37-421c-b670-297abe1beddc", "error": null}}'].
```

hamilton/function_modifiers/base.py
```diff
@@ -47,6 +51,14 @@ def validate(self, fn: Callable):
         """
         pass
 
+    def count_decorator_usage(self):
```
elijahbenizzy, line 54: Nit -- I don't really like this being a method of this class. It feels like the decorator shouldn't really care about it.

Rather, we should put this outside the class and wrap the `__call__` function in it (which gets called every time a decorator gets applied). Then we just pass it the global variable and it increments.

```python
def track_decorator_usage(__call__: Callable) -> Callable: 
    @functools.wraps(__call__)
    def replace__call__(*args, **kwargs):
        # do your counting/filtering/incrementing 
        return __call__(*args, **kwargs)
```

This way its less coupled logically and allows us to potentially reuse this. Global state is referenced outside the class.

hamilton/function_modifiers/base.py
```diff
@@ -47,6 +51,14 @@ def validate(self, fn: Callable):
...
+        global COUNTER
```
elijahbenizzy, line 55: Also this shouldn't have to be global as you assign it to a self variable in `decorator_counter` below

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+
+from .version import VERSION
+
+logger = logging.getLogger(__name__)
+
+STR_VERSION = ".".join([str(i) for i in VERSION])
+
+HOST = "https://app.posthog.com"
+TRACK_URL = f"{HOST}/capture/"  # https://posthog.com/docs/api/post-only-endpoints
+API_KEY = "phc_mZg8bkn3yvMxqvZKRlMlxjekFU5DFDdcdAsijJ2EH5e"
```
elijahbenizzy, line 32: OK, there has to be a better way of storing this. Someone will likely find and abuse, and our posthog bill (while free now) will go through the roof. 

Another reason for `hamilton --init`, which could allow us to generate an user/application-specific tracking key.

elijahbenizzy: The other approach is to ensure posthog is OK with rate limiting/validation on their part, but ughh its messy.

elijahbenizzy: Or we could operate a server that they can get a new API key from if telemetry is enabled... Might be worth talking to posthog and figure out how they recommend. Or some way of doing public encryption.

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+
+            g_telemetry_enabled = config.getboolean("DEFAULT", "telemetry_enabled")
+        except ValueError as e:
+            logger.debug(
+                "Unable to parse value for `telemetry_enabled` from config. " f"Encountered {e}"
+            )
+    if os.environ.get("HAMILTON_TELEMETRY_ENABLED") is not None:
+        env_value = os.environ.get("HAMILTON_TELEMETRY_ENABLED")
+        try:
+            g_telemetry_enabled = config._convert_to_boolean(env_value)
```
elijahbenizzy, line 88: Is there a cleaner approach than using the private method here?

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+    number_of_nodes: int,
+    number_of_modules: int,
+    number_of_config_items: int,
+    decorators_used: Dict[str, int],
+    graph_adapter_used: str,
+    result_builder_used: str,
+    driver_run_id: uuid.UUID,
+    error: Optional[str],
+):
+    """TODO: format into POSTHOG schema
```
elijahbenizzy, line 135: You already did this right?

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
+"""
+This module contains code that relates to sending Hamilton usage telemetry.
```
elijahbenizzy, line 2: More docstrings on this file

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+    :return: string to use for telemetry
+    """
+    try:
+        import traceback
+
+        te = traceback.TracebackException(exc_type, exc_value, exc_traceback, limit=-5)
+        sanitized_string = ""
+        for stack_item in te.stack:
+            stack_file_path = stack_item.filename.split(os.sep)
+            # take last 4 places only -- that's how deep hamilton is.
```
elijahbenizzy, line 250: Going to have to be convinced that this actually works...

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+            stack_file_path = stack_file_path[-4:]
+            if "hamilton" not in stack_file_path:
```
elijahbenizzy, line 252: Why not filter everything? As opposed to the last 4?

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+        try:
+            with open(DEFAULT_PROFILE, "w") as f:
+                config.write(f)
+        except Exception:
+            pass
+    return config
+
+
+config = _load_config()
+g_telemetry_enabled = True
```
elijahbenizzy, line 67: Relying a lot on global variables here...

Maybe a singleton class that stores everything and you can access/modify the fields?

hamilton/telemetry.py
```diff
@@ -0,0 +1,262 @@
...
+def create_end_event_json(
+    is_success: bool,
+    runtime_seconds: float,
+    number_of_outputs: int,
+    number_of_overrides: int,
+    number_of_inputs: int,
+    driver_run_id: uuid.UUID,
+    error: Optional[str],
+):
+    """TODO: format into POSTHOG schema
```
elijahbenizzy, line 179: This is done right?

hamilton/driver.py
```diff
@@ -20,7 +23,7 @@
     import graph
     import node
 else:
-    from . import base, graph, node
```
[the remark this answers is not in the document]
skrawcz: yeah I think for now custom drivers are out of scope.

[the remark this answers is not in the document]
skrawcz: actually, this code is here because of the testing script at the bottom of the file  -- and has nothing to do with custom drivers...

hamilton/driver.py
```diff
@@ -53,14 +56,63 @@ def __init__(
...
+            self.adapter = adapter
         except Exception as e:
+            error = telemetry.sanitize_error(*sys.exc_info())
             logger.error(SLACK_ERROR_MESSAGE)
             raise e
-        self.adapter = adapter
+        finally:
+            self.capture_constructor_telemetry(error, modules, config, adapter)
+
+    def capture_constructor_telemetry(
```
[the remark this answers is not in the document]
skrawcz: > Thoughts on having this live here versus in telemetry, or a new file hamilton_telemetry.py

Right now, the driver is the thing that weaves all the context together. I think I can put two helper functions in `telemetry` for testing/context purposes. Else the rest I think is fine to have here.

> adapter gets passed in but also saved to self.adapter

Is not guaranteed to exist. Thus better to pass it in.

> in fact pretty much all of these except config appear function_graph and could potentially be accessible.

Exactly "potentially". So being defensive here.

hamilton/driver.py
```diff
@@ -53,14 +56,63 @@ def __init__(
...
+        :return:
+        """
+        if telemetry.is_telemetry_enabled():
+            try:
+                if adapter.__module__.startswith("hamilton."):
+                    adapter_name = f"{adapter.__module__}.{adapter.__class__.__name__}"
+                else:
+                    adapter_name = "custom_adapter"
+                result_builder = "custom_builder"
+                if hasattr(adapter, "result_builder"):
```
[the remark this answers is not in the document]
skrawcz: If it's a custom adapter -- so being defensive :)

hamilton/driver.py
```diff
@@ -53,14 +56,63 @@ def __init__(
...
+                    if adapter.result_builder.__module__.startswith("hamilton."):
+                        result_builder = (
+                            f"{adapter.result_builder.__module__}."
+                            f"{adapter.result_builder.__class__.__name__}"
+                        )
+                payload = telemetry.create_start_event_json(
```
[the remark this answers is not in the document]
skrawcz: because the finally can be called in weird states.

hamilton/driver.py
```diff
@@ -53,14 +56,63 @@ def __init__(
...
+                    adapter_name,
+                    result_builder,  # check inheritance which one implements ResultMixin?
+                    self.driver_run_id,
+                    error,
+                )
+                telemetry.send_event_json(payload)
+            except Exception as e:
+                # we don't want this to fail at all!
+                if logger.isEnabledFor(logging.DEBUG):
+                    logger.debug(f"Telemetry caught error: {e}")
```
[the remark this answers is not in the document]
skrawcz: it's a catch 22 to send it. So I don't think this is recoverable.

hamilton/driver.py
```diff
@@ -133,12 +185,52 @@ def execute(
...
+            run_successful = False
             logger.error(SLACK_ERROR_MESSAGE)
+            error = telemetry.sanitize_error(*sys.exc_info())
             raise e
+        finally:
+            self.capture_execute_telemetry(
+                error, final_vars, inputs, overrides, run_successful, start_time
+            )
+
+    def capture_execute_telemetry(
```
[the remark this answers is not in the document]
skrawcz: will just make it duration.

hamilton/telemetry.py
skrawcz: from their docs: 
> Write-only means it can only create new events. It can't read events or any of your other data stored with PostHog, so it's safe to use in public apps.

hamilton/telemetry.py
skrawcz: that's what a module is effectively...

hamilton/telemetry.py
skrawcz: writing the code ourselves...

hamilton/function_modifiers/base.py
skrawcz: 🤔  I dont though....

hamilton/function_modifiers/base.py
```diff
@@ -47,6 +51,14 @@ def validate(self, fn: Callable):
...
+        if self.__module__.startswith("hamilton.function_modifiers"):
```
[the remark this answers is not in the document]
skrawcz: not sure that's necessary right now?

hamilton/telemetry.py
elijahbenizzy: IMO using a module to store global state is generally sloppy and should be limited

hamilton/driver.py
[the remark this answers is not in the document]
elijahbenizzy: Ahh good point

hamilton/telemetry.py
skrawcz: I refactored a bit...

README.md
```diff
@@ -257,6 +257,41 @@ If you are doing a lot of repetitive work, one might consider multiple cursors.
...
+# Usage analytics & data privacy
+By default, when using Hamilton, it collects anonymous usage data to help improve Hamilton and know where to apply development
+efforts.
+
+We capture two events: one when the `Driver` object is instantiated, and one when the `execute()` call on the `Driver` object completes.
+No user data or potentially sensitive information is or ever will be collected. The captured data is limited to:
+
+* Operating System and Python version
+* A persistent UUID to indentify the session, stored in ~/.hamilton.conf.
+* Error stack trace limited to Hamilton code, if one occurs.
```
elijahbenizzy, line 269: More info here about how you to build trust?

elijahbenizzy: Maybe a sample JSON payload to make it clear?

examples/async/fastapi_example.py
```diff
@@ -1,17 +1,23 @@
...
 
 @app.post("/execute")
 async def call(request: fastapi.Request) -> dict:
     """Handler for pipeline call"""
-    dr = h_async.AsyncDriver({}, async_module)
     input_data = {"request": request}
-    return await dr.raw_execute(["pipeline"], inputs=input_data)
+    # Can instantiate a driver within a request as well:
+    # dr = h_async.AsyncDriver({}, async_module, result_builder=base.DictResult())
+    result = await dr.execute(["pipeline"], inputs=input_data)
```
elijahbenizzy, line 19: Nit: why separate into two lines?

graph_adapter_tests/h_async/test_h_async.py
```diff
@@ -57,3 +59,25 @@ async def test_driver_end_to_end():
...
+        "another_async_func": 8,
+        "async_func_with_param": 4,
+        "external_input": 1,
+        "non_async_func_with_decorator": {"result_1": 9, "result_2": 5},
+        "result_1": 9,
+        "result_2": 5,
+        "simple_async_func": 2,
+        "simple_non_async_func": 7,
+    }
+    await asyncio.sleep(1)  # to ensure the last telemetry invocation finishes executing
```
elijahbenizzy, line 81: :/ classic "make-the-tests-slower-to-cover-for-side-effects-we-cant-test-for".

Fine for now but a while loop that burns through, checks, sleeps, then time's out. This might be what you want: https://docs.python.org/3.6/library/asyncio-sync.html#asyncio.Condition.wait_for.

hamilton/base.py
```diff
@@ -13,7 +13,10 @@
 import typing_inspect
 from pandas.core.indexes import extension as pd_extension
 
-from . import node
+try:
+    from . import node
```
elijahbenizzy, line 17: What?

hamilton/driver.py
```diff
@@ -133,12 +183,62 @@ def execute(
                 "display_graph=True is deprecated. It will be removed in the 2.0.0 release. "
                 "Please use visualize_execution()."
             )
+        start_time = time.time()
+        run_successful = True
+        error = None
         try:
             outputs = self.raw_execute(final_vars, overrides, display_graph, inputs=inputs)
-            return self.adapter.build_result(**outputs)
+            result = self.adapter.build_result(**outputs)
```
elijahbenizzy, line 191: Why two lines again?

hamilton/experimental/h_async.py
```diff
@@ -141,9 +144,72 @@ async def execute(
...
             logger.error(driver.SLACK_ERROR_MESSAGE)
+            error = telemetry.sanitize_error(*sys.exc_info())
             raise e
+        finally:
+            duration = time.time() - start_time
+            # ensure we can capture telemetry in async friendly way.
+            if telemetry.is_telemetry_enabled():
+
+                async def make_coroutine():
+                    self.capture_execute_telemetry(
```
elijahbenizzy, line 165: This doesn't return anything...

I'm pretty sure this will not actually use a coroutine as its not async. It'll just execute capture_execute_telelemetry.

For simplicity I'm not sure we should support async execute, just constructor.

hamilton/experimental/h_async.py
```diff
@@ -141,9 +144,72 @@ async def execute(
...
+
+        :param error: sanitized error string, if any.
+        :param modules: tuple of modules to build DAG from.
+        :param config: config to create the driver.
+        :param adapter: adapter class object.
+        """
+        if telemetry.is_telemetry_enabled():
+            try:
+                # check whether the event loop has been started yet or not
+                loop = asyncio.get_event_loop()
```
elijahbenizzy, line 195: Oh I see what this does -- I guess if it works then 🤷

hamilton/function_modifiers/base.py
```diff
@@ -18,6 +20,29 @@ def sanitize_function_name(name: str) -> str:
     return name[:last_dunder_index] if last_dunder_index != -1 else name
 
 
+COUNTER = collections.defaultdict(int)
```
elijahbenizzy, line 23: Nit -- should call this `DECORATOR_COUNTER` to be a little more readable.

elijahbenizzy (commented):
A few nits but looks good

hamilton/function_modifiers/base.py
skrawcz: good point.

README.md
skrawcz: I think I'll just point people to the code. That way it'll stay up to date.

examples/async/fastapi_example.py
skrawcz: that was me debugging things. no reason.

graph_adapter_tests/h_async/test_h_async.py
skrawcz: 🤔 not quite -- but I found another way that seems to work for 3.7+.

hamilton/base.py
skrawcz: to make `python driver.py` work, need to catch the import error.

hamilton/driver.py
skrawcz: likely debugging. but no reason.

hamilton/experimental/h_async.py
skrawcz: I'm pretty sure this is how you make something not async, async :)

README.md
skrawcz: which I point to below already.

elijahbenizzy (approved):

PR #256, opened 2022-12-23
Tag outputs

elijahbenizzy:
See https://github.com/stitchfix/hamilton/discussions/254


skrawcz:
Need a commit message that informs of design decisions please.

hamilton/function_modifiers/metadata.py
```diff
@@ -125,3 +125,49 @@ def validate(self, fn: Callable):
...
                 f"reserved as well: {self.RESERVED_TAG_NAMESPACES}"
             )
+
+
+class tag_nodes(base.SubDAGModifier):
+    def __init__(self, every_node_gets: Dict[str, str] = None, **tag_mapping: Dict[str, str]):
+        """Creates a tag_nodes decorator.
+
+        :param every_node_gets: Tags that every node gets -- this is the same as additionally using "@tag"
+        :param tag_mapping: Mapping of node name to tags -- this is aking to applying @tag to indicvidual nodes produced by the function
```
skrawcz, line 135:
```suggestion
        :param tag_mapping: Mapping of node name to tags -- this is akin to applying @tag to individual nodes produced by the function.
```

hamilton/function_modifiers/metadata.py
```diff
@@ -125,3 +125,49 @@ def validate(self, fn: Callable):
...
+        """
+        if every_node_gets is None:
+            every_node_gets = {}
+        self.every_node_gets = every_node_gets
+        self.tag_mapping = tag_mapping
+        pass
+
+    def _transform_node(self, node_: node.Node, config: Dict[str, Any], fn) -> node.Node:
```
skrawcz, line 143: doc string please

hamilton/function_modifiers/metadata.py
```diff
@@ -125,3 +125,49 @@ def validate(self, fn: Callable):
...
+        new_tags = node_.tags.copy()
+        new_tags.update(self.every_node_gets)  # apply tags that every node gets
+        new_tags.update(self.tag_mapping.get(node_.name, {}))
+        # This is slightly strange due to the the fact that transform_nodereturns a list...
```
skrawcz, line 147:
```suggestion
        # This is slightly strange due to the the fact that transform_node returns a list...
```

hamilton/function_modifiers/metadata.py
```diff
@@ -125,3 +125,49 @@ def validate(self, fn: Callable):
...
+            )
+        return [self._transform_node(node_, config, fn) for node_ in nodes]
+
+    @classmethod
+    def get_lifecycle_name(cls) -> str:
+        return "tag_nodes"
+
+    @classmethod
+    def allows_multiple(cls) -> bool:
+        return True  # why not? Order will matter here.
```
skrawcz, line 170: what does this mean? doc string to explain?

hamilton/function_modifiers/metadata.py
```diff
@@ -125,3 +125,49 @@ def validate(self, fn: Callable):
                 "The value can be anything. Note that the following top-level prefixes are "
                 f"reserved as well: {self.RESERVED_TAG_NAMESPACES}"
             )
+
+
+class tag_nodes(base.SubDAGModifier):
+    def __init__(self, every_node_gets: Dict[str, str] = None, **tag_mapping: Dict[str, str]):
+        """Creates a tag_nodes decorator.
+
+        :param every_node_gets: Tags that every node gets -- this is the same as additionally using "@tag"
```
skrawcz, line 134: why not just use `@tag` instead, and make this only allow a kwarg mapping?

e.g.
```python
@tag_nodes()
@tag()
```

skrawcz (commented):
Do we need `tag_nodes` to handle the "all case"? Why not just double up on decorators?

decorators.md
```diff
@@ -323,6 +325,50 @@ def intermediate_column() -> pd.Series:
 def final_column(intermediate_column: pd.Series) -> pd.Series:
     pass
 ```
+### @tag_outputs
+
+`tag_outputs` enables you to attach metadata to a function that outputs multiple nodes,
+and give different tag values to different items:
```
skrawcz, line 331:
```suggestion
and give different tag values to different outputs:
```
items isn't a term we use.

decorators.md
```diff
@@ -323,6 +325,50 @@ def intermediate_column() -> pd.Series:
...
+@tag_outputs(
+    public={'accessibility' : 'public'},
+    private={'accessibility' : 'private'})
+@extract_columns('public', 'private')
+def data_used_in_multiple_ways() -> pd.DataFrame:
+    return load_some_data(...)
+```
+
+In this case, the tag `accessibility` would have different values for the two nodes produced by the `data_used_in_multiple_ways`
+function -- `private` for `private` and `public` for `public`.
```
skrawcz, line 349:
```suggestion
@extract_columns('column_a', 'column_b')
def data_used_in_multiple_ways() -> pd.DataFrame:
    return load_some_data(...)
    ```

In this case, the tag `accessibility` would have different values for the two nodes produced by the `data_used_in_multiple_ways`
function -- `public` for `column_a` and `private` for `column_b`.
```
It's confusing to have columns and tags be named the same. So changed to make it clearer what is what.

skrawcz: note I don't know how to escape the backticks here in github...

decorators.md
```diff
@@ -305,7 +305,9 @@ To make this easier, we have a few more `@config` decorators:
       and resolves to
     `True` or `False`. You probably don't want to do this.
 
-## @tag
+## @tag and friends
+
+### @tag
 
 Allows you to attach metadata to a node (any node decorated with the function).
```
skrawcz, line 312:
```suggestion
Allows you to attach metadata to an output(s), i.e. all nodes generated by a function and its decorators.
```

decorators.md
```diff
@@ -323,6 +325,50 @@ def intermediate_column() -> pd.Series:
...
+@extract_columns('public', 'private')
+def data_used_in_multiple_ways() -> pd.DataFrame:
+    return load_some_data(...)
+```
+
+In this case, the tag `accessibility` would have different values for the two nodes produced by the `data_used_in_multiple_ways`
+function -- `private` for `private` and `public` for `public`.
+
+Note that you can use `@tag_outputs` to override `@tag` if you so desire. For that to work,
+you would put `tag_outputs` above `tag` (as it would be applied last).
```
skrawcz, line 352: Need to provide context for this.
```suggestion
A note on decorator precedence. If using `@tag` together with`@tag_outputs` on a function (you might want to do this because you use `@tag` to "tag" all nodes with a certain set of values, and `@tag_outputs` to "tag" specific outputs with specific values), they will be applied in order up from the function. So if you desire to override `@tag`, for that to work, you would put `tag_outputs` above `tag` (as it would be applied last).
```

decorators.md
```diff
@@ -323,6 +325,50 @@ def intermediate_column() -> pd.Series:
...
+    public={'accessibility' : 'public'},
+    private={'accessibility' : 'private', 'common_tag' : 'bar'})
+@tag(common_tag="foo")
+@extract_columns('public', 'private')
+def data_used_in_multiple_ways() -> pd.DataFrame:
+    return load_some_data(...)
+```
+
+In the case above, `common_tag` would resolve to `foo` for `public` and `bar` for `private`.
+Attempting an override in the reverse direction is currently undefined behavior.
```
skrawcz, line 371:
```suggestion
@extract_columns('column_a', 'column_b')
def data_used_in_multiple_ways() -> pd.DataFrame:
    return load_some_data(...)
    ```

In the case above, `common_tag` would resolve to `foo` for `column_a` and `bar` for `column_b`.
Attempting an override in the reverse direction is currently undefined behavior.
```

skrawcz (approved):
Just some docs tuning, otherwise looks good!

PR #263, opened 2023-01-01
Polars example

skrawcz:
Polars is gaining traction, we should have an example. This PR helps show an example that matches our hello world. It requires one adjust to the `extract_columns` decorator to function. Otherwise the user right now has to create their own build result function. Which seems fine for now, but it's something we could build a `sf-hamilton-polars` package for to house.

## Changes
1. prototypes change to extract_columns to handle providing the dataframe types.
2. adds example polars code.

## How I tested this
1. Tested this locally.

## Checklist

### DIFF
--- a/examples/polars/my_functions.py
+++ b/examples/polars/my_functions.py
@@ -3,7 +3,7 @@ import polars as pl
 from hamilton.function_modifiers import extract_columns
 
 
-@extract_columns("signups", "spend", df_type=pl.DataFrame, series_type=pl.Series)
+@extract_columns("signups", "spend")  # , df_type=pl.DataFrame, series_type=pl.Series)
 def base_df(base_df_location: str) -> pl.DataFrame:
     """Loads base dataframe of data.
 
--- a/examples/polars/my_functions.py
+++ b/examples/polars/my_functions.py
@@ -1,0 +1,48 @@
+import polars as pl
+
+from hamilton.function_modifiers import extract_columns
+
+
+@extract_columns("signups", "spend", df_type=pl.DataFrame, series_type=pl.Series)
+def base_df(base_df_location: str) -> pl.DataFrame:
+    """Loads base dataframe of data.
+
+    :param base_df_location: just showing that we could load this from a file...
+    :return:
+    """
+    return pl.DataFrame(
+        {
+            "signups": pl.Series([1, 10, 50, 100, 200, 400]),
+            "spend": pl.Series([10, 10, 20, 40, 40, 50]),
+        }
+    )
+
+
+def avg_3wk_spend(spend: pl.Series) -> pl.Series:
+    """Rolling 3 week average spend."""
+    return spend.rolling_mean(3)
+
+
+def spend_per_signup(spend: pl.Series, signups: pl.Series) -> pl.Series:
+    """The cost per signup in relation to spend."""
+    return spend / signups
+
+
+def spend_mean(spend: pl.Series) -> float:
+    """Shows function creating a scalar. In this case it computes the mean of the entire column."""
+    return spend.mean()
+
+
+def spend_zero_mean(spend: pl.Series, spend_mean: float) -> pl.Series:
+    """Shows function that takes a scalar. In this case to zero mean spend."""
+    return spend - spend_mean
+
+
+def spend_std_dev(spend: pl.Series) -> float:
+    """Function that computes the standard deviation of the spend column."""
+    return spend.std()
+
+
+def spend_zero_mean_unit_variance(spend_zero_mean: pl.Series, spend_std_dev: float) -> pl.Series:
+    """Function showing one way to make spend have zero mean and unit variance."""
+    return spend_zero_mean / spend_std_dev
--- a/hamilton/function_modifiers/expanders.py
+++ b/hamilton/function_modifiers/expanders.py
@@ -1,6 +1,6 @@
 import functools
 import inspect
-from typing import Any, Callable, Collection, Dict, Tuple, Union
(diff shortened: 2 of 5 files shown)


examples/polars/my_functions.py
```diff
@@ -0,0 +1,48 @@
+import polars as pl
+
+from hamilton.function_modifiers import extract_columns
+
+
+@extract_columns("signups", "spend", df_type=pl.DataFrame, series_type=pl.Series)
```
elijahbenizzy, line 6: Why do we need to type these if we can read it from the types below? Should be pretty clear that its a polars series from a polars dataframe...

hamilton/function_modifiers/expanders.py
```diff
@@ -308,14 +308,23 @@ class parameterized_inputs(parameterize_sources):
 
 
 class extract_columns(base.NodeExpander):
-    def __init__(self, *columns: Union[Tuple[str, str], str], fill_with: Any = None):
+    def __init__(
+        self,
+        *columns: Union[Tuple[str, str], str],
+        fill_with: Any = None,
+        df_type: Type = pd.DataFrame,
```
elijahbenizzy, line 315: OK, idea for the above. Take inspiration from singledispatch, and...

(1) Don't instantiate it on the dataframe type
(2) register additional types in plugins
(3) use dispatch on the node/DAG we get in

hamilton/function_modifiers/expanders.py
```diff
@@ -354,14 +365,18 @@ def expand_node(
...
             df_generated = fn(*args, **kwargs)
             if self.fill_with is not None:
                 for col in self.columns:
                     if col not in df_generated:
+                        # this wont work for polars for example
                         df_generated[col] = self.fill_with
             return df_generated
 
+        # manually add the type annotation
+        df_generator.__annotations__["return"] = self.df_type
```
elijahbenizzy, line 378: Good god

elijahbenizzy: Our design rule for this is we don't rely on typing from anonymous functions. These guys don't even need a return type cause we already specify it for the node. `.copy_with()` preserves the return type.

hamilton/function_modifiers/expanders.py
```diff
@@ -380,13 +395,16 @@ def extractor_fn(
                     )
                 return kwargs[node_.name][column_to_extract]
 
+            # manually add the type annotation
+            extractor_fn.__annotations__["return"] = self.series_type
```
elijahbenizzy, line 399: No!

elijahbenizzy (changes requested):
OK, all about these features but we should be a little smarter about how we do it -- this `extract_columns` approach is super brittle.

Some ideas:

1. Use singledispatch (or something like that) and register types
2. Define a set of operations on dataframes and implementations of them for each one
3. Have multiple decorators, using something like `applies_to` and then a dispatch-type approach.

Open to other ideas. Also this is making the assumption that we should know everything on instantiation of the decorator, which is wrong. We can actually derive it on calling of the decorator.

In fact, we may want to add a better validation step that validates when calling it -- our current one doesn't really do what we want.

hamilton/function_modifiers/expanders.py
skrawcz: yeah it's not used anyway. removing.

hamilton/function_modifiers/expanders.py
skrawcz: no used. removing.

hamilton/function_modifiers/expanders.py
skrawcz: 🤔  not sure how we can do `singledispatch` here without some major surgery on `extract_columns`; we want classes that aren't in the same module (don't want to add polars as a dependency to main package), where the only difference is really types in most cases, maybe a minor implementation change or two...

examples/polars/my_functions.py
skrawcz: We could read the dataframe type, sure, but we cannot read the `Series` type without hardcoding a mapping of dataframe to series somewhere and thus bringing in that dependency.... 🤔

elijahbenizzy: Yeah the polars plugin should do that... Otherwise the framework is saying "here's how we get from type X to Y" but not what "X" and "Y" are, and we only know how to get from "X" to "Y" in this one specific case...

hamilton/function_modifiers/expanders.py
elijahbenizzy: This can be solved with plugins and registration of extra implementations of decorators.

E.G. this is simple:
1) Polars plugin appends all its decorator implementations that are polars-specific
2) Polars plugin tells us the condition of when to use them pending the nodes we're modifying
3) We look to see what plugins are available and use the one that applies

I don't think extract_columns should have two different ways to call it? Its just going to get ugly.

skrawcz: I think we have several ways to implement a plugin.

It's mainly figuring out what should be abstracted. 

So I like the idea of abstracting the "dataframe" instead of the decorator. That way we use "black box" delegation, and the decorator is simple, one for all dataframes, it's the black box delegation that takes care of the specific implementation/behavior... For example, this is the code I was prototyping to see how it would look in my latest commit.

examples/polars/my_functions.py
skrawcz: yes

skrawcz:
TODOs:
 - [ ] determine dataframe interface
 - [ ] refactor into separate packages

skrawcz:
This was completed in #273

PR #273, opened 2023-01-12
Introduces *_extensions to extend functionality in hamilton

skrawcz:
Introduces *_extensions to extend functionality in hamilton

So we had a problem in the decorators (notably extract_columns)
that assumed pandas. We need to abstract this out so that other
dataframe libraries could be easily used with the decorator.

This is a commit that introduces a few changes to do that:

(1) We decided on a plugins module that houses two types of files:
*_extensions.py, and *_implementations.py. The *_extensions includes
code that the decorators and hamilton framework depends on.
The *_implementations will house specific implementations related to
that dataframe library for example, e.g. result builders. Users will reference these.
 Note we haven't created any _implementations.py files in this PR -- will do so in later ones.

(2) We decide that there are several functions that need to live in _extensions.py
that enable the hamilton framework to delegate/find the right thing to use.
This could be an object -- that one has to implement. But because it's already
in the module -- and we want things to be registered, I think this abstraction right
now is simpler -- the housing unit is the module that the functions are in -- don't need
to put them into an object right now.

(3) I decided trying to import the modules and having them determine how to error
would be the best way to determine if the rest of the module import should proceed.

## Changes
 - adds plugins module
 - adds to hamilton.base.py  to house registering extension options
 - adds extensions for polars, pandas, and pandas on pyspark
 - adds polars example and tweaks pandas on pyspark one to exercise extract_columns

## How I tested this
Ran code locally.

## Notes
This will help solve a user issue.

## Checklist


hamilton/base.py
```diff
@@ -21,6 +23,47 @@
...
+        if series_type is not None:
+            return series_type
+    raise NotImplementedError(f"Unknown dataframe type: {dataframe_type}. Cannot get Column type.")
+
+
+plugins_modules = ["pandas", "polars", "pyspark"]
+
+for plugin_module in plugins_modules:
+    try:
+        importlib.import_module(f"hamilton.plugins.{plugin_module}_extensions")
```
elijahbenizzy, line 62: check for functions here to ensure "interface"/

elijahbenizzy: and the logger.info("detected X support, loading extension for it")

elijahbenizzy: logger.debug("did not detect...")

hamilton/base.py
```diff
@@ -20,6 +22,81 @@
...
+    mod = importlib.import_module(f"hamilton.plugins.{plugin_module}_extensions")
+    assert hasattr(mod, "DATAFRAME_TYPE"), "Error extension mission DATAFRAME_TYPE"
+    assert hasattr(mod, "COLUMN_TYPE"), "Error extension mission COLUMN_TYPE"
+    assert hasattr(
+        mod, f"get_column_{plugin_module}"
+    ), f"Error extension mission get_column_{plugin_module}"
+    assert hasattr(
+        mod, f"fill_with_scalar_{plugin_module}"
+    ), f"Error extension mission fill_with_scalar_{plugin_module}"
+    global DF_TYPE_AND_COLUMN_TYPES
```
elijahbenizzy, line 84: So I think this works -- only weird thing is that we do this differently than we do registry for single dispatch.

- Modules are responsible for registering their own implementations
- Types are added centrally...

Might be nice to have modules add their own types?

hamilton/base.py
```diff
@@ -20,6 +22,81 @@
...
+        "dataframe_type": mod.DATAFRAME_TYPE,
+        "column_type": mod.COLUMN_TYPE,
+    }
+    logger.info(f"Detected {plugin_module} and successfully loaded Hamilton extensions.")
+
+
+plugins_modules = ["pandas", "polars", "pyspark_pandas"]
+for plugin_module in plugins_modules:
+    try:
+        _load_extension(plugin_module)
```
elijahbenizzy, line 95: Isn't this a circular dependency?
- plugin_module depends on base to register
- we import plugin_module here
- this happens at import time...

hamilton/function_modifiers/expanders.py
```diff
@@ -335,9 +335,12 @@ def validate(self, fn: Callable):
...
         """
         output_type = inspect.signature(fn).return_annotation
-        if not issubclass(output_type, pd.DataFrame):
+        try:
+            hamilton_base.get_column_type_from_df_type(output_type)
+        except NotImplementedError:
             raise base.InvalidDecoratorException(
-                f"For extracting columns, output type must be pandas dataframe, not: {output_type}"
+                # TODO: capture was dataframe libraries are supported and print here.
+                f"Error {fn} does not output a type we know about. Is it a dataframe type we support?"
```
elijahbenizzy, line 343: Maybe a link to open an issue/reach out?

hamilton/plugins/pandas_extensions.py
```diff
@@ -0,0 +1,23 @@
...
+
+try:
+    import pandas as pd
+except ImportError:
+    raise NotImplementedError("Pandas is not installed.")
+
+from hamilton import base
+
+# Required constants!
+DATAFRAME_TYPE = pd.DataFrame
```
elijahbenizzy, line 11: If we want we can statically validate these extensions to make sure we don't change anything -- parse an AST and look for these nodes...

hamilton/plugins/polars_implementations.py
```diff
@@ -0,0 +1,27 @@
+from typing import Dict
+
+import polars as pl
+
+from hamilton import base
+
+
+class PolarsDataFrameResult(base.ResultMixin):
```
elijahbenizzy, line 8: Thought -- if we had a general "concat" operation or "join" operation we could implement a more generic result builder. We now have the ability to register operations...

elijahbenizzy (commented):
LGTM -- some qs about circular dependencies but I imagine you've tested it and I'm missing something...

hamilton/base.py
skrawcz: yep

hamilton/base.py
skrawcz: yep

hamilton/function_modifiers/expanders.py
skrawcz: They should get the pointer to slack when this occurs.

hamilton/plugins/polars_implementations.py
skrawcz: Maybe. But polars doesn't have an index -- so 🤷.

PR #275, opened 2023-01-13
Uses the base python graph adapter for type-checking in experimental

elijahbenizzy:
graph adapters

Before we were doing it in a custom way -- this broke when a user wanted to use generics. We still don't have amazing generic/type comparison, but this should be able to clean it up.

## Checklist

### DIFF
--- a/graph_adapter_tests/h_dask/test_h_dask.py
+++ b/graph_adapter_tests/h_dask/test_h_dask.py
@@ -52,12 +52,15 @@ def test_smoke_screen_module(client):
         "pessimistic_net_acquisition_cost",
         "neutral_net_acquisition_cost",
         "optimistic_net_acquisition_cost",
+        "series_with_start_date_end_date",
     ]
     df = dr.execute(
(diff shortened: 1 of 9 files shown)


elijahbenizzy:
Worth testing? Probably only with integration tests -- all the functionality is unit-tested.

skrawcz:
This may or may not be the right approach.

What we really need is a test case to reproduce the issue that was seen :)

elijahbenizzy:
Yeah this was a quick educated guess. IMO this is the right thing to do, we just need to validate that it's the best way to solve the problem. IMO inheritance is a little cleaner for this sort of thing too...

skrawcz:
> Yeah this was a quick educated guess. IMO this is the right thing to do, we just need to validate that it's the best way to solve the problem. Inheritance is probably a little cleaner for this sort of thing too...

Actually yeah I think that should work. The point is that these adapters should fudge whatever types they need to, but otherwise delegating to super feels right overall for them.

skrawcz (approved):

skrawcz (approved):

PR #283, opened 2023-01-23
Adds ability to pass in a callable to get the function name from for …

skrawcz:
…execute

I found myself wanting to click directly to outputs being created to inspect them. So to short cut that, directly referencing them is a way I can do that and have my IDE jump there when navigating code.

The downside is that this solution doesn't handle functions annotated with decorators that create nodes, e.g. @parameterize*, or @config* . I think that's okay, because it's probably clearer to have a "string" name here to indicate that something is changing/being produced at DAG creation time, and thus a reference to an exact function is likely a misnomer...

Adds unit test for this.

Example usage:
```python
import data_loaders, transforms, model_pipeline
...
dr = driver.Driver(config, data_loaders, transforms, model_pipeline)  
dr.visualize_execution([model_pipeline.kaggle_submission_df], "./kaggle_submission_df.dot.png", {})
kaggle_submission_df: pd.DataFrame = dr.execute([model_pipeline.kaggle_submission_df])
```
The list can be a union of strings or callables. We right now just grab the name of the callable.

## Changes
- execute can now take in a "callable" that we take the name from.
- adds unit test

## How I tested this
 - locally
 - via unit tests

## Notes
 - this will also make it simpler for IDEs to find references and potentially rename things when function names change.

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -9,7 +9,7 @@ import uuid
 from dataclasses import dataclass, field
 from datetime import datetime
 from types import ModuleType
-from typing import Any, Callable, Collection, Dict, List, Optional, Set, Tuple
+from typing import Any, Callable, Collection, Dict, List, Optional, Set, Tuple, Union
 
 import pandas as pd
 
@@ -203,14 +203,14 @@ class Driver(object):
 
     def execute(
         self,
-        final_vars: List[str],
+        final_vars: List[Union[str, Callable]],
         overrides: Dict[str, Any] = None,
         display_graph: bool = False,
         inputs: Dict[str, Any] = None,
     ) -> Any:
         """Executes computation.
 
-        :param final_vars: the final list of variables we want to compute.
+        :param final_vars: the final list of outputs we want to compute.
         :param overrides: values that will override "nodes" in the DAG.
         :param display_graph: DEPRECATED. Whether we want to display the graph being computed.
         :param inputs: Runtime inputs to the DAG.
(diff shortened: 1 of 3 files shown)


hamilton/driver.py
```diff
@@ -238,9 +240,37 @@ def execute(
...
+        errors = []
+        module_set = {_module.__name__ for _module in self.graph_modules}
+        for final_var in final_vars:
+            if isinstance(final_var, str):
+                _final_vars.append(final_var)
+            elif isinstance(final_var, Callable):
+                if final_var.__module__ in module_set:
+                    _final_vars.append(final_var.__name__)
+                else:
+                    errors.append(
```
elijahbenizzy, line 262: Edge-case *might* be with create_temporary_module, but (a) I think it'll still work and (b) not worth worrying about

hamilton/driver.py
```diff
@@ -238,9 +240,37 @@ def execute(
...
+                        f"Function {final_var.__module__}.{final_var.__name__} is a function not in a "
```
elijahbenizzy, line 263: This is only guarding for the case of using the function with the same name from a different (not included) module, correct? Otherwise if its from a module that wasn't specified and doesn't exist, it won't be an appropriate variable..

elijahbenizzy (approved):
LGTM 🚢 🇮🇹

hamilton/driver.py
skrawcz: I don't thinks so looking at the code -- but either way not worth worrying about:

The temporary module does do this:
```python
fn.__module__ = module.__name__
```
so I think it'll work.

hamilton/driver.py
skrawcz: yup, and yup.

hamilton/driver.py
elijahbenizzy: Yep that's what I was thinking as well

PR #286, opened 2023-01-25
Changes visualize driver functions to return graphviz object

skrawcz:
This change is backwards compatible. It returns and wires through returning the graphviz object created. That way in a notebook, it can render the DAG for you.

It also makes saving to a file optional -- you just need to pass in None explicitly. Adds test for it.

The hello world example is also updated with a real notebook and shows Thierry Jean's approach of using the autoreload extensions.

## Changes
 - returns graphviz object from display functions
 - changes hello world to include a proper notebook

## How I tested this
 - via notebook
 - unit tests

## Checklist


skrawcz:
Screen Shot 2023-01-24 at 9 54 16 PM

hamilton/driver.py
```diff
@@ -359,7 +359,7 @@ def list_available_variables(self) -> List[Variable]:
     @capture_function_usage
     def display_all_functions(
         self, output_file_path: str, render_kwargs: dict = None, graphviz_kwargs: dict = None
-    ):
+    ) -> Optional["graphviz.Digraph"]:  # noqa F821
```
elijahbenizzy, line 362: Why optional?

hamilton/graph.py
```diff
@@ -243,12 +243,12 @@ def display(
         output_file_path: str = "test-output/graph.gv",
         render_kwargs: dict = None,
         graphviz_kwargs: dict = None,
-    ):
+    ) -> Optional["graphviz.Digraph"]:  # noqa F821
```
elijahbenizzy, line 246: Not sure this one should be optional -- feel like its always nice to return and it won't break anything (at worst will make people's notebooks look better).

tests/test_graph.py
```diff
@@ -530,6 +530,20 @@ def test_function_graph_display():
...
+    fg = graph.FunctionGraph(tests.resources.dummy_functions, config={"b": 1, "c": 2})
+    defined_nodes = set()
+    user_nodes = set()
+    for n in fg.get_nodes():
+        if n.user_defined:
+            user_nodes.add(n)
+        else:
+            defined_nodes.add(n)
+    digraph = fg.display(defined_nodes, user_nodes, None)
+    assert digraph is not None
```
elijahbenizzy, line 544: Should assert instance/something else here to be a little more specific IMO

elijahbenizzy (commented):
LGTM -- a few minor points

hamilton/driver.py
skrawcz: because if graphviz is not installed you will not get anything back.

hamilton/graph.py
skrawcz: because if graphviz is not installed it wont return an object.

tests/test_graph.py
skrawcz: checking the graph isn't valuable because there's already a test for that. But sure I can check the instance type.

elijahbenizzy (approved):

PR #305, opened 2023-02-08
Adds GeoPandas dataframe support

skrawcz:
We missed adding this when we removed the "pandas" like dataframe assumption.

## Changes
 - Adds geopandas plugin

## How I tested this
 - Runs locally on something that uses geopandas

## Checklist


elijahbenizzy (approved):
LGTM

PR #311, opened 2023-02-14
Adds basic scenario 1 for feature engineering

skrawcz:
This example shows how you can use the same feature definitions in Hamilton in an offline setting and use them in an online setting.

Assumptions:
 - the API request can provide the same raw data that training provides.
 - if you have aggregation features, you need to store the training result for them, and provide them to the online side.

## Checklist


examples/feature_engineering/README.md
```diff
@@ -0,0 +1,32 @@
...
+that result in subtle bugs and hard to maintain code.
+
+With this example we show how you can use Hamilton to:
+
+1. write a feature once.
+2. leverage the feature code anywhere that python runs. e.g. in batch and online.
+3. show how to modularize components so that if you have values cached in a feature store,
+you can inject those values into your feature computation needs.
+
+# Scenarios
```
elijahbenizzy, line 14: Might be nice to share the common goal (E.G. what are we actually trying to predict/what pipeline are we trying to build?

examples/feature_engineering/scenario_1/constants.py
```diff
@@ -0,0 +1,11 @@
+model_x_features = [
```
elijahbenizzy, line 1: Comments as to why we're storing model features this way would be helpful

examples/feature_engineering/scenario_1/fastapi_server.py
```diff
@@ -0,0 +1,132 @@
...
+    return {
+        "age_mean": 33.0,
+        "age_std_dev": 12.0,
+    }
+
+
+def fake_model(df: pd.DataFrame) -> pd.Series:
+    """Function to simulate a model"""
+    # do some transformation.
+    return df.sum()  # this is nonsensical but provides a single number.
```
elijahbenizzy, line 40: For actually releasing it, let's actually have a model here... 

What does the outerbounds/metaflow example do? We could just load their model and referenced it.

examples/feature_engineering/scenario_1/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### offline_loader.py
+Contains logic to load raw data. Here it's a flat file, but it could be going
+to a database, etc.
+
+#### features.py
+The feature transform logic that takes raw data and transforms it into features. It contains some runtime
+dataquality checks using Pandera.
+
+Important not, there are two aggregations features defined: `age_mean` and `age_std_dev`, that are computed on the
```
elijahbenizzy, line 32: In truth, we only need the combined one, right? As in, that's the one feature that's used? `age_zero_mean_unit_variance`?

examples/feature_engineering/scenario_1/fastapi_server.py
elijahbenizzy: But if you do want faked data, just stick it in another file:

```
return data.load_model()...
return data.load_invariant_features()...
```

We want this to look as much like an API as they would be used to...

Another idea: the features could be tagged with `invariant` versus `variant` -- allowing us to know which ones to inject.

elijahbenizzy (commented):
Good start -- I don't think this is going to be clear to most people who haven't really dug into this. A few thoughts:

1. We can clarify the wording/make it crisper to specify why this is a problem, how its normally done, and why hamilton alleviates this
2. We can give more context about what we're doing here/why its in an online context
3. We can root on tooling that might be familiar to them. While loading fake models/whatnot makes sense, I think its going to confuse the users. So either load from a model/feature store they're used to, or (more likely) abstract it away and make it very clear that it could be implemented in many different ways.

This stuff is natural to us as we've been building online/batch inference/training tooling for years, but I think this will be extremely complex to most people out there, and fall flat. Hamilton is simple enough and makes this easy enough that this is a good chance to capture market share, but to do so we need to really hammer home a pattern and a motivation.

examples/feature_engineering/scenario_1/README.md
skrawcz: They help compute that feature, given a person's age. No way around not computing it.

examples/feature_engineering/README.md
skrawcz: 🤔 sure.

skrawcz:
> Good start -- I don't think this is going to be clear to most people who haven't really dug into this. A few thoughts:
> 
> 1. We can clarify the wording/make it crisper to specify why this is a problem, how its normally done, and why hamilton alleviates this
> 2. We can give more context about what we're doing here/why its in an online context
> 3. We can root on tooling that might be familiar to them. While loading fake models/whatnot makes sense, I think its going to confuse the users. So either load from a model/feature store they're used to, or (more likely) abstract it away and make it very clear that it could be implemented in many different ways.
> 
> This stuff is natural to us as we've been building online/batch inference/training tooling for years, but I think this will be extremely complex to most people out there, and fall flat. Hamilton is simple enough and makes this easy enough that this is a good chance to capture market share, but to do so we need to really hammer home a pattern and a motivation.

That's the point of the scenarios, there is no one size fits all. That is, show the simplest possible thing, then one where there is a feature store, etc.  

Will add more to motivation -- and draw some pictures.

examples/feature_engineering/scenario_1/README.md
elijahbenizzy: Ahh, yeah, of course. I think this is a little tricky -- might be worth digging into/explaining more

examples/feature_engineering/scenario_1/fastapi_server.py
skrawcz: I think there is a bit of overhead to more files -- more things to grok.

Otherwise tagging is another concept to talk about and add -- not sure it's worth it, but maybe? 🤔

elijahbenizzy: Yeah, I could go either way

skrawcz: added tags.

PR #317, opened 2023-02-18
Subdag decorator

elijahbenizzy:
Makes `@subdag` usable. Things to think about/fix:
- [ ] Doesn't work with extract_column, might want to fix
- [ ] Docs need a bit of love
- [ ] Should the `load_from` parameter be a param or a list of inputs?

## Checklist

### DIFF
--- a/decorators.md
+++ b/decorators.md
@@ -399,21 +399,26 @@ Note that you can also specify custom decorators using the `@check_output_custom
 
 See [data_quality](data_quality.md) for more information on available validators and how to build custom ones.
 
-## @reuse_functions
+## @subdag
 
-Currently located under `experimental` -- looking for feedback!
-
-The `@reuse_functions` decorator enables you to rerun components of your DAG with varying parameters. Note that this is immensely powerful -- if we
+The `@subdag` decorator enables you to rerun components of your DAG with varying parameters. Note that this is immensely powerful -- if we
 draw analogies from Hamilton to standard procedural programming paradigms, we might have the following correspondence:
 
 - `config.when` + friends -- `if/else` statements
 - `parameterize`/`extract_columns` -- `for` loop
 - `does` -- effectively macros
-And so on. `@reuse_functions`takes this one step further.
-- `@reuse_functions` -- subroutine definition
+And so on. `@subdag` takes this one step further.
+- `@subdag` -- subroutine definition
 E.G. take a certain set of nodes, and run them with specified parameters.
 
-If you're confused as to why you need this decorator, you should probably stop reading (you most likely don't need it). If this solves a pain point you've had, then continue...
+Why might you want to use this? Let's take a look at some examples:
+
+1. You have a feature engineering pipeline that you want to run on multiple datasets. If its exactly the same, this is perfect. If not, this
+works perfectly as well, you just have to utilize different functions in each or the `config.when` + `config` parameter to rerun it.
+2. You want to train multiple models in the same DAG that share some logic (in features or training) -- this allows you to reuse and continually add more.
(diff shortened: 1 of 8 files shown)


examples/reusing_functions/reusable_subdags.py
```diff
@@ -1,9 +1,8 @@
 import pandas as pd
 import unique_users
 
-from hamilton.experimental.decorators import reuse
-from hamilton.experimental.decorators.reuse import reuse_functions
-from hamilton.function_modifiers import value
+from hamilton.function_modifiers import reuse, value
+from hamilton.function_modifiers.reuse import reuse_functions
```
skrawcz, line 5: reuse and reuse_functions doesn't seem right here?

hamilton/function_modifiers/recursive.py
```diff
@@ -229,55 +215,70 @@ def fn(
...
-            new_nodes.append(
-                create_identity_node(
-                    from_=assign_namespace(from_node, self.namespace),
-                    name=to_node,
-                    typ=nodes_by_name[from_node_namespaced].type,
-                    namespace=(),
-                )
-            )
-        return nodes + new_nodes
+    # def _add_output_nodes(self, nodes: List[node.Node]) -> List[node.Node]:
```
skrawcz, line 218: why commented out?

hamilton/storage/loader.py
```diff
@@ -0,0 +1,12 @@
+import abc
```
skrawcz, line 1: committed by mistake?

tests/resources/reuse_subdag.py
```diff
@@ -35,48 +36,56 @@ def f(c: int, b: int, e: int) -> int:
...
+    config={"op": "add"},
 )
-def subdag_4() -> MultiOutput(e_4=int, f_4=int):
-    pass
+def v4(e: int, f: int) -> dict:
+    return {"e_4": e, "f_4": f}
+
+
+# This is a little messy because we can't currently use @extract_fields
+# in conjunction with @subdag. So we just extract afterwards..
```
skrawcz, line 76: what's the limitation here? what about extract_columns?

skrawcz (changes requested):

decorators.md
```diff
@@ -432,28 +432,27 @@ def unique_users(filtered_interactions: pd.DataFrame, grain: str) -> pd.Series:
...
-@reuse_functions(
-    with_inputs={"grain": value("day")},
-    namespace="daily_users_US",
-    outputs={"unique_users": "unique_users_daily_US"},
-    with_config={"region": "US"},
-    load_from=[unique_users, interactions_filtered],
+@subdag(
+    unique_users, interactions_filtered,
+    inputs={"grain": value("day")},
+    config={"region": "US"},
```
skrawcz, line 438: should probably have functions that have drivers within them to help highlight the functionality? i.e. instead of using a driver within a function and having 0 visibility, you can write these functions...

tests/resources/reuse_subdag.py
elijahbenizzy: That doesn't work either. The limitation is that we currently don't allow it -- `extract_columns/fields` has a requirement that it only works for situations in which there is just one output.

In reality, it needs to work for situations in which the subdag on which it operates has one "sink" node. Didn't want to do it, so I'll add an issue after this.

skrawcz: Cool -- yeah file the issue, sounds good.

skrawcz (approved):
Seems to work. Looks good to go.

Issue #46, opened 2022-02-03
Capturing wide to long transformations of entire dataframes using Hamilton


latlan1:
**Is your feature request related to a problem? Please describe.**
Is there a way to transform a dataframe from wide to long (or vice versa) using Hamilton to track this transformation? I concluded no since all of the input/output columns would need to be specified, which could be a lot of typing. 

**Describe the solution you'd like**
It would nice if I could define a function that accepts df_wide and outputs df_long with pd.melt.

**Describe alternatives you've considered**
I performed the melt operation outside of Hamilton so this operation is not directly captured through the DAG.

skrawcz:
A recap of our discord conversation:

# what you can implement now
If you implement it as a function right now, you will lose expressiveness as the transform takes in a fixed set of columns. You would not be able to pick and choose what went into dynamically when using hamilton driver.
E.g.
```python
@extract_columns([index_value, name, value])
def wide_to_long(col1, col2, ...) -> pd.DataFrame:
    df = pd.DataFrame(... col1, ... col2 ...).
    return df.melt(...)
```

# Future Option 1
Introduce the ability to tell Hamilton how to "combine" results. We'd pass in an object of some type to tell the driver what to do.
e.g.
```python
dr = driver.Driver(config_or_data, modules, adapter=LongFormatDF())
long_df = dr.execute(['col1', ..., 'col3'])
# more hamilton stuff
dr2 = driver.Driver({'col1': long_df.col1, ...}, modules)
df2 = dr2.execute(['colX', ..., 'colY'])
```
This would not enable a single pass through, but it would enable you to chain computation in a way that enables changing what columns are transformed to long format easily.

# Future Option 2
We enable the ability to configure the inputs to a function via some configuration, and in the driver you'd be able to specify the final result, and have the DAG know what inputs are required.
e.g. 
```python
@extract_columns([index_value, name, value])
@variable_args
def wide_to_long(**variable_args) -> pd.DataFrame:
    df = pd.DataFrame(**variable_args).
    return df.melt(...)

dr = driver.Driver({..., wide_to_long_args=['col1', 'col2', ...]}, modules)
df_final = dr.execute(...)
```

# Directional thinking
We're pretty close to enabling option 1 with #36. I think that it forces a nice decoupling of concerns in a way -- it's more obvious what's happening, versus people having to understand configuration that's passed in via option 2. But I could see a future where we enable both options actually...

elijahbenizzy:
[...]

PR #321, opened 2023-02-19
Adds code to make concatenating dataframes columnwise in the result builder default

skrawcz:
Why this change? because if you are doing indicator variable stuff, maybe you don't want to be explicit about them, instead you want to concatentate the columns created without knowing about them. Seems like a reasonable pattern on first glance.

This code is backwards compatible with current users, and just enables one to have dataframes in the output.
 All it does is that it flattens the dataframe, and then ensures there's no duplicate column definitions. I added tests for this for this case -- this is all on the assumption that the indexes can be merged/used for the outer join (which is the current
behavior).

For the future, we probably want some flexibility with respect
to "joins" and how things are stitched together since you could
envisage some use cases where people want to inner join between
things --- but until there's a common pattern/need, I think
we should just keep thinking about what should be first class within Hamilton,
and what we get people to do themselves.

## Changes
 - base pandas dataframe result builder 

## How I tested this
 - unit tests

## Notes
 - should be 100% backwards compatible with current users, but now instead of erroring on a dataframe, it will columnwise concatenate it.

## Checklist

### DIFF
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -190,11 +190,12 @@ class PandasDataFrameResult(ResultMixin):
         :return: A dataframe with the outputs.
         """
         flattened_outputs = {}
-        outputs_present = set()
         for name, output in outputs.items():
             if isinstance(output, pd.DataFrame):
-                df_columns = set(output.columns)
-                column_intersection = outputs_present.intersection(df_columns)
+                df_columns = list(output.columns)
+                column_intersection = [
+                    column for column in df_columns if column in flattened_outputs
+                ]
                 if column_intersection:
                     raise ValueError(
                         f"Dataframe {name} contains columns {column_intersection} that already exist in the output. "
@@ -206,7 +207,6 @@ class PandasDataFrameResult(ResultMixin):
                     )
                 df_dict = output.to_dict(orient="series")
                 flattened_outputs.update(df_dict)
-                outputs_present = outputs_present.union(df_columns)
             elif isinstance(output, pd.Series):
                 if name in flattened_outputs:
                     raise ValueError(
@@ -214,7 +214,6 @@ class PandasDataFrameResult(ResultMixin):
                         f"Please rename the series to avoid this error, or determine from where the initial series is "
                         f"being added; it may be coming from a dataframe that is being unpacked."
                     )
-                outputs_present.add(name)
                 flattened_outputs[name] = output
             else:
                 if name in flattened_outputs:
@@ -223,7 +222,6 @@ class PandasDataFrameResult(ResultMixin):
                         f"Please rename this output to avoid this error, or determine from where the initial value is "
                         f"being added; it may be coming from a dataframe that is being unpacked."
                     )
-                outputs_present.add(name)
                 flattened_outputs[name] = output
         return pd.DataFrame(flattened_outputs)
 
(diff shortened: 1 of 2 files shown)


skrawcz:
This follows conversation in https://github.com/stitchfix/hamilton/issues/46 where I created a result builder for a user; thanks to @Arnechos for the push/idea.

hamilton/base.py
```diff
@@ -169,8 +169,63 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
...
+        :param outputs: The outputs to build the dataframe from.
+        :return: A dataframe with the outputs.
+        """
+        flattened_outputs = {}
+        outputs_present = set()
+        for name, output in outputs.items():
+            if isinstance(output, pd.DataFrame):
+                df_columns = set(output.columns)
+                column_intersection = outputs_present.intersection(df_columns)
+                if column_intersection:
```
elijahbenizzy, line 198: The other options is to namespace them with the name of the dataframe... Although I think this is cleaner and more intuitive.

hamilton/base.py
```diff
@@ -169,8 +169,63 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
...
+        The behavior of pd.Dataframe(outputs) is that it will do an outer join based on indexes of the Series passed in.
+        To handle dataframes, we unpack the dataframe into a dict of series, check to ensure that no columns are
+        redefined in a rolling fashion going in order of the outputs requested. This then results in an "enlarged"
+        outputs dict that is then passed to pd.Dataframe(outputs) to get the final dataframe.
+
+        :param outputs: The outputs to build the dataframe from.
+        :return: A dataframe with the outputs.
+        """
+        flattened_outputs = {}
+        outputs_present = set()
```
elijahbenizzy, line 193: Shouldn't you be able to tell which outputs are present based on the var `flattened_outputs`? Rather than having to store a set?

elijahbenizzy (commented):
A minor point. Also let's add to documentation. Otherwise 👍

hamilton/base.py
skrawcz: yeah I'm optimizing here. Since I'd be looping over a set of columns, so I thought I'd just make a set to make it simpler...

skrawcz: i.e. for each column, check if in dict. else just do a set operation.

hamilton/base.py
skrawcz: yeah could use the name of the dataframe for that, but thought we should wait for a user to complain before adding that.

hamilton/base.py
```diff
@@ -169,8 +169,63 @@ def build_result(**outputs: Dict[str, Any]) -> pd.DataFrame:
             # If we're dealing with all values that don't have any "index" that could be created
             # (i.e. scalars, objects) coerce the output to a single-row, multi-column dataframe.
             return pd.DataFrame([outputs])
+        #
+        contains_df = any(isinstance(value, pd.DataFrame) for value in outputs.values())
+        if contains_df:
+            # build the dataframe from the outputs
+            return PandasDataFrameResult.build_dataframe_with_dataframes(outputs)
+        # don't do anything special if dataframes aren't in the output.
+        return pd.DataFrame(outputs)  # this does an implicit outer join based on index.
```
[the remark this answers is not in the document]
skrawcz: This is that, no?

hamilton/base.py
elijahbenizzy: Yeah I'd just make it the `dict` not the set. One less thing to keep track of and its the same speed.

hamilton/base.py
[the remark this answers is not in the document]
elijahbenizzy: Ahh, misread

elijahbenizzy (approved):

hamilton/base.py
skrawcz: done.

PR #322, opened 2023-02-20
Tweaks logging when concatenating Dataframes

skrawcz:
We should push for a convention to help make it easy to debug those columns and find the code if there is an issue. I don't hard code the convention here because I don't want to preclude things. Otherwise an annoying warning is probably good for now -- and could be removed/moved to debug later.

## Changes
- base.py

## How I tested this
- locally

## Checklist

### DIFF
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -189,33 +189,25 @@ class PandasDataFrameResult(ResultMixin):
         :param outputs: The outputs to build the dataframe from.
         :return: A dataframe with the outputs.
         """
+
+        def get_output_name(output_name: str, column_name: str) -> str:
+            """Add function prefix to columns.
+            Note this means that they stop being valid python identifiers due to the `.` in the string.
+            """
+            return f"{output_name}.{column_name}"
+
         flattened_outputs = {}
         for name, output in outputs.items():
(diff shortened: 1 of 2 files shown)


hamilton/base.py
```diff
@@ -192,20 +192,16 @@ def build_dataframe_with_dataframes(outputs: Dict[str, Any]) -> pd.DataFrame:
...
-                    )
                 if logger.isEnabledFor(logging.DEBUG):
                     logger.debug(
-                        f"Unpacking dataframe {name} into dict of series with columns {df_columns}."
+                        f"Unpacking dataframe {name} into dict of series with columns {list(output.columns)}."
                     )
-                df_dict = output.to_dict(orient="series")
+                # add function prefix to columns -- note this means that they stop being valid python identifiers
+                # due to the `.` in the string.
+                df_dict = {
```
elijahbenizzy, line 201: Stick the naming piece into a function `get_output_column_name(df_name: str, col_name: str)`? Or something like that. Makes it clearer/separates the logic so readers down the line will figure it out.

skrawcz: sure. done.

elijahbenizzy (approved):

